PR #1, opened 2019-08-14
Adding protobuf definitions for the 1st version of the API

wk8:
All these are subject to changes down the line as we start implementing
each RPC call; but this gives a draft of what we expect the alpha API to
look like.

Signed-off-by: Jean Rouge <rougej+github@gmail.com>


msau42:
cc @jingxu97

api/smb.proto
```diff
@@ -0,0 +1,59 @@
...
+    // All special characters allowed by Windows in path names will be allowed
+    // except for restrictions noted below. For details, please check:
+    // https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+    //
+    // Restrictions:
+    // If given an absolute path, it needs to be under kubelet-csi-plugins-path,
+    // and specified with drive letter prefix: "X:\".
+    // If given a relative path, it will be understood as being relative to
+    // kubelet-csi-plugins-path.
+    // Needs to already exist on host.
```
pjh: There are two ways to use the `New-SmbGlobalMapping` command for mounting SMB shares on the container host to prepare for mounting them as container volumes:

1) `New-SmbGlobalMapping -RemotePath "\\$ServerName\$ShareName" -LocalPath X: -Credential cred` . This seems to be what `MountSMBShareRequest` is targeting, but in this case it might make more sense to just accept a drive letter here (which does not need to exist on the host I think?), and then use the other proxy APIs to create any child directories.

2) `New-SmbGlobalMapping -RemotePath "\\$ServerName\$ShareName" -Credential $cred` (no `-LocalPath`) followed by a symlink to e.g. `C:\smbshare`. This is the approach used by the existing kubernetes [SMB mount package](https://github.com/kubernetes/kubernetes/blob/2036ee5ddeac235c0c39c705ad22cfa19bad309e/pkg/util/mount/mount_windows.go#L109). `MountSMBShareRequest` should allow this (or a separate request type should be created) since in our experience we've encountered strange permission errors when mounting SMB shares created using approach 1) to containers, whereas shares created using approach 2) can be mounted to containers as expected.

cc @almos98

api/file_system.proto
```diff
@@ -0,0 +1,156 @@
...
+    // except for restrictions noted below. For details, please check:
+    // https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+    //
+    // Restrictions:
+    // If given an absolute path, it needs to be under kubelet-pod-path,
+    // and specified with drive letter prefix: "X:\".
+    // If given a relative path, it will be understood as being relative to
+    // kubelet-pod-path.
+    // Needs to already exist on host.
+    // UNC paths of the form "\\server\share\path\file" are not allowed.
```
pjh: See my comment below and the [mount_windows package](https://github.com/kubernetes/kubernetes/blob/2036ee5ddeac235c0c39c705ad22cfa19bad309e/pkg/util/mount/mount_windows.go#L109) - I think symlinks with remote server shares as the source might be needed in order to mount SMB shares to containers.

api/smb.proto
```diff
@@ -0,0 +1,59 @@
+syntax = "proto3";
+
+package api;
+
+import "errors.proto";
+
+service SmbCSIProxyService {
+    // MountSMBShare mounts a remote share over SMB on the host at the requested global staging path.
+    rpc MountSMBShare(MountSMBShareRequest) returns (MountSMBShareResponse) {}
+}
```
pjh: `UnmountSMBShare` is needed too.

api/errors.proto
```diff
@@ -0,0 +1,31 @@
...
+// CommandError details errors yielded by cmdlet calls.
+message CmdletError {
+    // Name of the cmdlet that errored out.
+    string cmdlet_name = 1;
+
+    // Error code that got returned.
+    uint32 code = 2;
+
+    // Human-readable error message - can be empty.
+    string message = 4;
```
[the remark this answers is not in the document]
wk8: 👍 good catch, thanks!

api/smb.proto
wk8: I think approach 2) is also what we're targeting here; does 1) allow for some use cases that 2) doesn't?

api/file_system.proto
wk8: I think I'm missing something here, but as long as `MountSMBShareRequest` already symlinks the remote server share to within the proxy's working directories, why would users need to symlink to outside of it?

Again, apologies if I'm missing something :)

api/smb.proto
wk8: Indeed...! Adding

wk8: As well as `UnmountISCSILun`...

api/smb.proto
pjh: FYI the limitation with approach 1) is described at https://github.com/moby/moby/issues/37863.

I guess I'm confused by these comments for this method:
```
    // If given an absolute path, it needs to be under kubelet-csi-plugins-path,
    // and specified with drive letter prefix: "X:\".
    // If given a relative path, it will be understood as being relative to
    // kubelet-csi-plugins-path.
    // Needs to already exist on host.
```

The X:\ letter and "needs to already exist on host" are confusing. For approach 2) this method could either just create the mapping and then let the driver call a separate method to create the symlink, or this method could do both but I think the path would need to *not* already exist on the host (and the drive letter would typically be C:\).

msau42:
/retest

trying to pickup new prow job

api/iscsi.proto
```diff
@@ -0,0 +1,364 @@
+syntax = "proto3";
+
+package api;
+
+import "errors.proto";
+
+service IscsiCSIProxyService {
```
ddebroy: Can we also have [AddIScsiConnection](https://docs.microsoft.com/en-us/windows/win32/api/iscsidsc/nf-iscsidsc-addiscsiconnectionw) here?

api/general_storage.proto
```diff
@@ -0,0 +1,189 @@
...
+    map <string, DiskLocation> disk_locations = 4;
+}
+
+message ListDiskIDsRequest {
+    // Intentionally empty
+}
+
+message DiskIDs {
+    // list of Disk IDs of ASCII characters associated with disk device.
+    repeated string IDs = 1;
```
ddebroy: Each DiskID should also have a "type" member (potentially an enum) and "metadata" member (potentially a string) that will report the type of the disk ID we will support querying for and additional metadata associated with the ID. 
For Page 83 IDs, "type" can be: Page0x83 and "metadata" can the VPD Descriptor Type.
For Windows IDs, "type" can be: WindowsOS and "metadata" can be empty (as it will always point to Get-PhysicalDisk UniqueId)

api/service.proto
```diff
@@ -0,0 +1,103 @@
...
+    rpc FormatVolume(FormatVolumeRequest) returns (FormatVolumeResponse) {}
+
+    // ListDiskLocations returns locations <Adapter, Bus, Target, LUN ID> of all disk devices enumerated by Windows.
+    rpc ListDiskLocations(ListDiskLocationsRequest) returns (ListDiskLocationsResponse) {}
+
+    // ListDiskIDs returns all IDs (from IOCTL_STORAGE_QUERY_PROPERTY) of all disk devices enumerated by Windows.
+    rpc ListDiskIDs(ListDiskIDsRequest) returns (ListDiskIDsResponse) {}
+
+    // ListDiskVolumeMappings returns a map of all disk devices and volumes GUIDs
+    rpc ListDiskVolumeMappings(ListDiskVolumeMappingsRequest) returns (ListDiskVolumeMappingsResponse) {}
```
jingxu97: What are the use cases of these three list functions? Is there some doc or link you can share?

api/service.proto
```diff
@@ -0,0 +1,103 @@
...
+
+    // ResizeVolume performs resizing of the partition and file system for a block based volume
+    rpc ResizeVolume(ResizeVolumeRequest) returns (ResizeVolumeResponse) {}
+
+    // DismountVolume gracefully dismounts a volume
+    rpc DismountVolume(DismountVolumeRequest) returns (DismountVolumeResponse) {}
```
jingxu97: Is there a MountVolume call?

ddebroy: Good catch. We need something along the lines of a `MountBlockVolume` (just like `MountSMBShare`). `MountBlockVolumeRequest` will need to comprise of:
```
string volume_device_guid 
string host_path
bool readonly
```
`MountBlockVolumeResponse` will comprise of `int32 error_code`.

api/service.proto
ddebroy: `ListDiskLocations` or `ListDiskIDs` will be used by a block based CSI node driver to identify a disk device published to a node. 

`ListDiskIDs` maps to invoking `scsi-id` in Linux as done at https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/gcepd/gce_util.go#L231 as well as what `Get-GcePdName` achieves internally when invoked through: https://github.com/kubernetes/kubernetes/blob/17997ab7161736b23359480c774415bda3877130/pkg/volume/gcepd/attacher.go#L180. All SCSI Page 83 IDs (along with other IDs) for all disk devices will be returned here.

`ListDiskLocations` maps to invoking the `Get-Disk` cmdlet and looking at standard disk fields as done at https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/azure_dd/azure_common_windows.go#L42

`ListDiskVolumeMappings` is required to obtain a mapping of volume device GUIDs as enumerated by Windows to the disk device objects backing the volumes. This information will be used by a block based CSI driver to enumerate the volume devices hosted on a disk (published by the CSI attacher). The volume device GUID for a volume can be passed to `MountVolume` to mount the volume at the desired global staging path.

api/smb.proto
wk8: Fair points, changing :)

api/iscsi.proto
wk8: Adding

api/general_storage.proto
wk8: 👍 adding

api/service.proto
wk8: Adding comments

api/service.proto
wk8: 👍 adding

wk8:
/hold
will have to refactor when/if https://github.com/kubernetes-csi/csi-proxy/pull/7 merges, to account for the structure that latter PR imposes on API definition files.

api/file_system.proto
```diff
@@ -0,0 +1,156 @@
...
+    // Directories will be created with Read and Write privileges of the Windows
+    // User account under which csi-proxy is started (typically LocalSystem).
+    //
+    // Restrictions:
+    // If given an absolute path, it needs to be under kubelet-csi-plugins-path
+    // or kubelet-pod-path - based on context, and be specified with drive letter
+    // prefix: e.g., "X:\".
+    // If given a relative path, it will be understood as being relative to either
+    // kubelet-csi-plugins-path or kubelet-pod-path - again based on context.
+    // Cannot already exist on host.
```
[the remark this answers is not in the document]
jingxu97: When returning errors, can caller differentiate different types of errors, e.g., Dir already exists?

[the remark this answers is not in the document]
wk8: Yes - all of these API calls map to either a powershell cmdlet or a Windows API procedure, and in either case failures will return either a `WindowsApiError` or a `CmdletError` (from https://github.com/kubernetes-csi/csi-proxy/pull/1/files#diff-6048105abef6df2d3f269061114b54c3), so callers can differentiate that way.

This also keeps in line with the general philosophy of having as little logic as possible in this repo.

api/file_system.proto
```diff
@@ -0,0 +1,156 @@
...
+    // Path needs to be a directory that is empty.
+    // Maximum path length will be capped to 260 characters (MAX_PATH).
+    string path = 1;
+
+    // Context of the path creation used for path prefix validation
+    PathContext context = 2;
+}
+
+message RmdirResponse {
+    bool success = 1;
```
jingxu97: do we really need success bool? If no error, does it mean RmDir succeeded?

ddebroy: maybe the error code can be directly used to surface error. 0 = ERROR_SUCCESS based on MSDN.

api/iscsi.proto
```diff
@@ -0,0 +1,349 @@
...
```
pjh: I see both `Iscsi` and `IScsi` below. Can you make the capitalization consistent?

PR #3, opened 2019-08-15
Sync release tools

msau42:
Import all of csi-release-tools to this repo

### DIFF
--- a/release-tools/build.make
+++ b/release-tools/build.make
@@ -1,0 +1,178 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+.PHONY: build-% build container-% container push-% push clean test
+
+# A space-separated list of all commands in the repository, must be
+# set in main Makefile of a repository.
+# CMDS=
+
+# This is the default. It can be overridden in the main Makefile after
+# including build.make.
+REGISTRY_NAME=quay.io/k8scsi
+
+# Revision that gets built into each binary via the main.version
+# string. Uses the `git describe` output based on the most recent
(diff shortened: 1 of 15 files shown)


release-tools/build.make
```diff
@@ -0,0 +1,178 @@
...
+else
+TESTARGS =
+endif
+
+# Specific packages can be excluded from each of the tests below by setting the *_FILTER_CMD variables
+# to something like "| grep -v 'github.com/kubernetes-csi/project/pkg/foobar'". See usage below.
+
+build-%:
+	mkdir -p bin
+	CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$* ./cmd/$*
```
ddebroy: We will add the Windows target here in a later PR.

release-tools/build.make
```diff
@@ -0,0 +1,178 @@
...
+
+container-%: build-%
```
ddebroy: Given`csi-proxy` is going to be a host binary, I think this will need to change to a model more aligned with what is done for kubelet/kubeproxy binary releases rather than a container build. How about removing the `container` and `push` target below for now?

release-tools/travis.yml
```diff
@@ -0,0 +1,19 @@
+language: go
```
ddebroy: Should we consider appveyor here given the focus on Windows? Example usage in containerd: https://github.com/containerd/containerd/blob/master/.appveyor.yml

release-tools/build.make
msau42: all the files under release-tools are supposed to be shared and consistent across all kubernetes-csi repos, so we can't really remove things, only add new windows rules on the side.

release-tools/travis.yml
msau42: I would be open to it.  Can we get someone dedicated to setup the build/test infra for this project?

As a side note, we're also trying to move completely to prow and not use travisci anymore, so we may need to sync up with them on Windows support

ddebroy: @wk8 will be looking into it. He is applying for membership to kubernetes-csi for easy tagging.

ddebroy:
Let's get this in for now as the baseline. We can add the Windows targets and adjustments in a follow up PR.

release-tools/travis.yml
wk8: @msau42 : no travis also means no 3rd-party CI at all? No appveyor either?

msau42: We should check with test-infra on their plans to support Windows in prow.  If it's not anytime soon, then I am open to using a 3rd party

PR #11, opened 2019-10-04
FileSystem APIs

ddebroy:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Initial structure and APIs for supporting host file system calls

Fixes #

Very much WIP - just the initial structure

```release-note
NONE
```


ddebroy:
cc @wk8

client/api/errors.proto
```diff
@@ -13,3 +13,21 @@ message CmdletError {
...
+    // Name of the Windows API procedure that returned an error.
+    string procedure_name = 1;
+
+    // Error code that got returned. For additional details of a code please refer 
+    // to https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes
+    uint32 code = 2;
+
+    // Windows API's documentation lists hexadecimal error codes; this is the hexadecimal
+    // representation of the error code. For additional details, please refer
+    // to https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/705fb797-2175-4a90-b5a3-3918024b10b8
```
wk8: Might be worth linking to https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/18d8fbe8-a967-4f1c-ae50-99ca8e491d2d too?

wk8: Might be worth linking to https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/18d8fbe8-a967-4f1c-ae50-99ca8e491d2d too?

client/api/errors.proto
```diff
@@ -13,3 +13,21 @@ message CmdletError {
...
+message WindowsApiError {
+    // Name of the Windows API procedure that returned an error.
+    string procedure_name = 1;
+
+    // Error code that got returned. For additional details of a code please refer 
+    // to https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes
+    uint32 code = 2;
+
+    // Windows API's documentation lists hexadecimal error codes; this is the hexadecimal
+    // representation of the error code. For additional details, please refer
```
wk8: Could mention that'll be a string of the form `0x[0-9A-F]{8}` ?

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -4,25 +4,156 @@ package v1alpha1;
 
 import "github.com/kubernetes-csi/csi-proxy/client/api/errors.proto";
 
-// FIXME: this is just an almost empty service, not actually implemented; it's just
-// here as an example of what API group and versions will look like. It will actually
-// be implemented in a later patch.
-
 service FileSystem {
-    // PathExists checks if the given path exists on the host.
+    // PathExists checks if the parameter path exists on the host filesystem.
```
wk8: nit: `host's filesystem`?

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -4,25 +4,156 @@ package v1alpha1;
...
     rpc PathExists(PathExistsRequest) returns (PathExistsResponse) {}
+
+    // Mkdir creates a directory at the requested path on the host filesystem.
```
wk8: Same nit (and same for other occurrences below)

cmd/server/main.go
```diff
@@ -3,9 +3,17 @@ package main
 import (
 	"github.com/kubernetes-csi/csi-proxy/internal/server"
 	filesystem "github.com/kubernetes-csi/csi-proxy/internal/server/filesystem"
+	flag "github.com/spf13/pflag"
+)
+
+var (
+	kubeletCSIPluginsPath = flag.String("kubelet-csi-plugins-path", "C:\\var\\lib\\kubelet\\plugins", "Absolute path of the Kubelet plugin directory in the host file system")
```
wk8: Nit: using backticks for strings containing backslashes is more readable IMO.

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -4,25 +4,156 @@ package v1alpha1;
...
+
+    // Context of the path parameter.
+    // This is used to determine the root for relative path parameters
+    PathContext context = 2;
 }
 
 message PathExistsResponse {
-    bool success = 1;
+    // The Windows API error code observed (if any)
+    api.WindowsApiError api_error = 2;
```
wk8: The way I imagined it, `CmdletError` would be used when the RPC call mapped to (and was being run as) a powershell cmdlet, and `WindowsApiError` when the RPC call mapped to a Windows API procedure (e.g. https://docs.microsoft.com/en-us/windows/win32/api/iscsidsc/nf-iscsidsc-addiscsiconnectionw).

Not sure `PathExists` maps to either of these though; presumably that would be just calling `os.Stat` in go? So maybe a simple `string error = 1` would be enough here?

wk8: And in any case this should be field # 1 , no one out there is running the mock API that's currently in master :)

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -4,25 +4,156 @@ package v1alpha1;
...
+    // Context of the path parameter.
+    // This is used to [1] determine the root for relative path parameters
+    // or [2] validate prefix for absolute paths (indicated by a drive letter 
+    // prefix: e.g. "C:\")
+    PathContext context = 2;
+}
+
+message MkdirResponse {
+    // The Windows API error code observed (if any)
+    api.WindowsApiError api_error = 1;
```
wk8: Same remark as above, not sure this should be a `WindowsApiError`; maybe a `CmdletError` if the implementation shells out to `New-Item`, or just a string error message if using `os.MkdirAll`?

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -4,25 +4,156 @@ package v1alpha1;
...
+    // Context of the path creation used for path prefix validation
+    // This is used to [1] determine the root for relative path parameters
+    // or [2] validate prefix for absolute paths (indicated by a drive letter 
+    // prefix: e.g. "C:\")
+    PathContext context = 2;
+}
+
+message RmdirResponse {
+    // The Windows API error code observed (if any)
+    api.WindowsApiError api_error = 1;
```
wk8: Same remark.

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -4,25 +4,156 @@ package v1alpha1;
...
+    // target_path needs to exist as a directory in the host that is empty.
+    // target_path cannot be a symbolic link.
+    // Maximum path length will be capped to 260 characters.
+    string target_path = 2;
+}
 
-    bool exists = 4;
+message LinkPathResponse {
+    // The Windows API error code observed (if any)
+    api.WindowsApiError api_error = 1;
```
wk8: And same :)

client/api/errors.proto
jingxu97: what is the use case of this hexadecimal_code?

client/api/filesystem/v1alpha1/api.proto
jingxu97: if the requested path exists...?

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -4,25 +4,156 @@ package v1alpha1;
...
-    // PathExists checks if the given path exists on the host.
+    // PathExists checks if the parameter path exists on the host filesystem.
     rpc PathExists(PathExistsRequest) returns (PathExistsResponse) {}
+
+    // Mkdir creates a directory at the requested path on the host filesystem.
+    rpc Mkdir(MkdirRequest) returns (MkdirResponse) {}
+
+    // Rmdir removes the directory at the requested path on the host filesystem.
+    // This may be used for unlinking a symlink created through LinkPath.
+    rpc Rmdir(RmdirRequest) returns (RmdirResponse) {}
```
jingxu97: if the dir has content in it, it will delete the content too?

ddebroy: I think we can add a parameter to differentiate the two options - remove if non-empty vs remove iff directory is empty.

jingxu97: what is the expected behavior for non-empty dir?

ddebroy:
Execution of unit tests:
```
/go/src/github.com/kubernetes-csi/csi-proxy # GO111MODULE=on go test github.com/kubernetes-csi/csi-proxy/internal/server/filesystem
ok  	github.com/kubernetes-csi/csi-proxy/internal/server/filesystem	0.007s
```

client/api/filesystem/v1alpha1/api.proto
ddebroy: Should be an error.

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+		Exists: exists,
+	}, nil
+}
+
+func (s *Server) Mkdir(ctx context.Context, request *internal.MkdirRequest, version apiversion.Version) (*internal.MkdirResponse, error) {
+	err := s.validatePathWindows(request.Context, request.Path)
+	if err != nil {
+		return &internal.MkdirResponse{
+			Error: err.Error(),
+		}, nil
```
ddebroy: returning `nil` for the overall `err` is intentional here to distinguish between actual GRPC layer failures or failures in dispatching the API request to the right API handler vs errors in executing the API by the handler. We can optionally  use https://godoc.org/google.golang.org/grpc/status to directly return rich error data through API handlers (which is the pattern is CSI plugins).

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -2,27 +2,156 @@ syntax = "proto3";
...
+    // User account under which csi-proxy is started (typically LocalSystem).
+    //
+    // Restrictions:
+    // If an absolute path (indicated by a drive letter prefix: e.g. "C:\") is passed,
+    // depending on the context parameter of this function, the path prefix needs 
+    // to match the paths specified either as kubelet-csi-plugins-path
+    // or as kubelet-pod-path parameters of csi-proxy.
+    // If a relative path is passed, depending on the context parameter of this 
+    // function, the path will be considered relative to the path specified either as 
+    // kubelet-csi-plugins-path or as kubelet-pod-path parameters of csi-proxy.
```
jingxu97: In what case kubelet-csi-plugins-path will be used? and what case for kubelet-pod-path?

internal/os/filesystem/api.go
```diff
@@ -0,0 +1,50 @@
...
+	return true, err
+}
+
+func (APIImplementor) Mkdir(path string) error {
+	err := os.MkdirAll(path, 0755)
+	return err
+}
+
+func (APIImplementor) Rmdir(path string) error {
+	err := os.RemoveAll(path)
```
jingxu97: RemoveAll  removes path and any children it contains. But I thought we only remove empty dir which should be os.Remove(path)?

internal/os/filesystem/api.go
```diff
@@ -0,0 +1,50 @@
...
+	return err
+}
+
+func (APIImplementor) LinkPath(tgt string, src string) error {
+	var err error
+	if runtime.GOOS == "windows" {
+		_, err = exec.Command("cmd", "/c", "mklink", "/D", tgt, src).CombinedOutput()
```
jingxu97: Will os.Symlink() work here?
I do see an issue opened related to this https://github.com/golang/go/issues/22874
not sure what is the conclusion

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
-			CmdletName: "dummy",
-			Code:       12,
-			Message:    "hey there " + request.Path,
-		},
+		Error:  err.Error(),
+		Exists: exists,
+	}, nil
+}
+
+func (s *Server) Mkdir(ctx context.Context, request *internal.MkdirRequest, version apiversion.Version) (*internal.MkdirResponse, error) {
```
jingxu97: the passing parameter ctx context.Context is not used?

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+		}, nil
+	}
+
+	return &internal.MkdirResponse{
+		Error: "",
+	}, nil
+}
+
+func (s *Server) Rmdir(ctx context.Context, request *internal.RmdirRequest, version apiversion.Version) (*internal.RmdirResponse, error) {
+	err := s.validatePathWindows(request.Context, request.Path)
```
jingxu97: the passing parameter ctx context.Context is not used?

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+			Error: err.Error(),
+		}, nil
+	}
+	return &internal.RmdirResponse{
+		Error: "",
+	}, nil
+}
+
+func (s *Server) LinkPath(ctx context.Context, request *internal.LinkPathRequest, version apiversion.Version) (*internal.LinkPathResponse, error) {
+	err := s.validatePathWindows(internal.CONTAINER, request.SourcePath)
```
jingxu97: why here to use internal.CONTAINER directly instead of request.Context?

client/api/filesystem/v1alpha1/api.proto
ddebroy: I will add that in the comments. For `NodeStageVolume` calls, kubelet-csi-plugins-path is expected to be used to create `C:\var\lib\kubelet\plugins\kubernetes.io\csi\pv\<pv-name>\globalmount`. For `NodePublishVolume` calls,, kubelet-pod-path is expected to be used to create `C:\var\lib\kubelet\pods\<pod-uuid>\volumes\kubernetes.io~csi\<pvc-name>\mount`.

internal/server/filesystem/server.go
ddebroy: Correct. `ctx` is not very relevant for quick operations like the FS ones and therefore not used. A rough parallel is `ctx` passed to all CSI API calls but drivers typically ignore it when servicing the API calls in the [Identity set](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/330efe6a568a25eb8430218ca568fe7f710b5622/pkg/gce-pd-csi-driver/identity.go#L30).

internal/server/filesystem/server.go
ddebroy: Both are intentional. For `ctx`, please see [above](https://github.com/kubernetes-csi/csi-proxy/pull/11#discussion_r339925937). For `apiversion.Version`, it is not required right now given this is the v1alpha1 - the initial implementation. It will become relevant as we add more versions of the APIs and there arises a need to differentiate between API versions that are sending the calls to the underlying/internal implementation after parameter conversions.

internal/server/filesystem/server.go
ddebroy: In general we want to constrain the operations within FS contexts as much as possible (as commented in [KEP](https://github.com/kubernetes/enhancements/pull/1184#discussion_r308935813)). For Mkdir/Rmdir, we need to allow both as explained [above](https://github.com/kubernetes-csi/csi-proxy/pull/11#discussion_r339923451). LinkPath however is only needed for linking the node-global staging paths under PLUGIN context to pod-local staging paths under CONTAINER context.

client/api/filesystem/v1alpha1/api.proto
wk8: I like the idea of the parameter. Are you planning on implementing this in this PR? Or follow up?

internal/os/filesystem/api.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (APIImplementor) PathExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true, nil
+	}
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return true, err
```
wk8: Why true?

internal/os/filesystem/api.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (APIImplementor) Mkdir(path string) error {
+	err := os.MkdirAll(path, 0755)
+	return err
```
wk8: Nit: why not just `return os.MkdirAll` ?

internal/os/filesystem/api.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (APIImplementor) Rmdir(path string) error {
+	err := os.RemoveAll(path)
+	return err
+}
+
+func (APIImplementor) LinkPath(tgt string, src string) error {
+	var err error
+	if runtime.GOOS == "windows" {
```
wk8: This is never meant to run anywhere else though, so not sure I see the value of that `if`?

internal/server/filesystem/server_test.go
```diff
@@ -0,0 +1,238 @@
...
+}
+func (fakeFileSystemAPI) Rmdir(path string) error {
+	return nil
+}
+func (fakeFileSystemAPI) LinkPath(tgt string, src string) error {
+	return nil
+}
+
+func TestMkdirWindows(t *testing.T) {
+	v1alpha1, nil := apiversion.NewVersion("v1alpha1")
```
wk8: Wasn't aware golang had some pattern matching ;)

This only compiles because you shadow `nil`, then use that on L114. Please remove?

internal/server/filesystem/server_test.go
```diff
@@ -0,0 +1,238 @@
...
+			t.Errorf("Expected error but Mkdir returned a nil error")
+		}
+		if !tc.expectError && mkdirResponse.Error != "" {
+			t.Errorf("Expected no errors but Mkdir returned error: %s", mkdirResponse.Error)
+		}
+	}
+}
+
+func TestRmdirWindows(t *testing.T) {
+	v1alpha1, nil := apiversion.NewVersion("v1alpha1")
```
wk8: Same

internal/server/filesystem/server_test.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		t.Fatalf("FileSystem Server could not be initialized for testing: %v", err)
+	}
+	for _, tc := range testCases {
+		t.Logf("test case: %s", tc.name)
+		req := &internal.MkdirRequest {
+			Path: tc.path,
+			Context: tc.pathCtx,
+		}
+		mkdirResponse, _ := srv.Mkdir(context.TODO(), req, tc.version)
```
wk8: This got me wondering if it might be better for the internal server struct to define their callbacks either with or without (if they don't need it) version argument, and amend the autogen accordingly... Any take on this?

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+
+type API interface {
+	PathExists(path string) (bool, error)
+	Mkdir(path string) error
+	Rmdir(path string) error
+	LinkPath(tgt string, src string) error
+}
+
+func NewServer(hostOS string, kubeletCSIPluginsPath string, kubeletPodPath string, hostAPI API) (*Server, error) {
+	if hostOS != "windows" {
```
wk8: I'm sorry, I'm not sure I get the idea behind this; could you please elaborate a mind what you have in mind? It seems it never gets called with anything other than windows, even for tests, which I thought was the idea in the 1st place?

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+	}
+	return &Server{
+		kubeletCSIPluginsPath: kubeletCSIPluginsPath,
+		kubeletPodPath:        kubeletPodPath,
+		hostAPI:               hostAPI,
+	}, nil
+}
+
+func isDriveLetterWindows(c uint8) bool {
+	return ('a' <= c && c <= 'z' || 'A' <= c && c <= 'Z')
```
wk8: Nit: parens not needed?

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+	if len(path) < 2 {
+		return false
+	}
+	if path[0] == '\\' && path[1] == '\\' {
+		return true
+	}
+	return false
+}
+
+func isAbsWindows(path string) bool {
```
wk8: Nit: I personally would find this more readable as a regex, but your call :)

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
 
-// Server is the internal server for the filesytem API group.
-type Server struct{}
+type Server struct {
+	kubeletCSIPluginsPath string
+	kubeletPodPath        string
+	hostAPI               API
+}
+
+const MAX_PATH_LENGTH_WINDOWS = 260
```
wk8: Nit: this will certainly be used in many other places in this repo, might be worth putting it into an `internal/utils.go` file or similar?

internal/server/server.go
```diff
@@ -10,20 +10,21 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/kubernetes-csi/csi-proxy/client"
+	srvtypes "github.com/kubernetes-csi/csi-proxy/internal/server/types"
```
wk8: Need to amend the auto-gen to account for this, will do as this PR gets merged.

internal/os/filesystem/api.go
ddebroy: good catch. Will return `false` here.

internal/os/filesystem/api.go
ddebroy: os.Symlink() appears to work pretty well. I will switch to it. Thanks for the suggestion!

internal/server/filesystem/server.go
ddebroy: I agree this is useless. It's an artifact of something else I was trying and will remove.

internal/server/filesystem/server_test.go
ddebroy: heh - can't believe that compiled. Will remove :-)

client/api/filesystem/v1alpha1/api.proto
ddebroy: I will ammend in this PR.

internal/server/server.go
ddebroy: Thanks! Note that the auto-gen will also need to handle https://github.com/kubernetes-csi/csi-proxy/pull/11/files#diff-d81281b59794bf1f85ac717a4410af56.

internal/os/filesystem/api.go
ddebroy: I will parameterize this and call the appropriate API.

internal/server/filesystem/server_test.go
ddebroy: Good point. I guess in most scenarios we will not need the version argument. Probably fine to keep it in for now till at least v1beta1 and see?

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -2,27 +2,156 @@ syntax = "proto3";
...
+    // This may be used for unlinking a symlink created through LinkPath
+    rpc Rmdir(RmdirRequest) returns (RmdirResponse) {}
+
+    // LinkPath creates a local directory symbolic link between a source path 
+    // and target path in the host's filesystem
+    rpc LinkPath(LinkPathRequest) returns (LinkPathResponse) {}
+}
+
+// Context of the paths used for path prefix validation
+enum PathContext {
```
kkmsft: Wouldn't we need a NONE type to represent absolute paths which are independent of the kubelet-csi-plugins-path/kubelet-pod-path ?

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+		return fmt.Errorf("Invalid PathContext: %v", pathCtx)
+	}
+
+	pathlen := len(path)
+
+	if pathlen > MAX_PATH_LENGTH_WINDOWS {
+		return fmt.Errorf("Path length %d exceeds maximum characters: %d", pathlen, MAX_PATH_LENGTH_WINDOWS)
+	}
+
+	if len(path) > 0 && (path[0] == '\\') {
```
kkmsft: nit: we could use `pathlen` here instead of len(path)

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,227 @@ package filesystem
...
+
+	if isAbsWindows(path) && !strings.HasPrefix(path, prefix) {
+		return fmt.Errorf("Absolute path: %s is not within context path: %s", path, prefix)
+	}
+
+	return nil
+}
+
+func (s *Server) abs(pathCtx internal.PathContext, path string) (string, error) {
+	if isAbsWindows(path) {
```
kkmsft: Shouldn't we let the caller decide what type the path needs to interpreted as - either ABSOLUTE path or PLUGIN or CONTAINER path? Currently, we seem to be have a mode in which we assume that the user intended that its absolute path by looking at the path and then fall through to other modes of PLUGIN and CONTAINER. I am thinking we could either have: 
-  Callers performs the append and passes the full path in all cases.
or 
- Have another mode internal.ABSOLUTE where we can explicitly know the caller intent. We can do the validations we do in isAbsWindows.

@ddebroy - WDYT ?

internal/os/filesystem/api.go
kkmsft: Wouldn't we want to use async mode in these cases - link I am guessing is not time consuming so not exactly relevant. But from the code flow from the top of the server, shouldn't we have a context which can be passed on to server to cancel the operation, which should be passed up to here and then we can use exec.Start/exec.Wait and exec.Process.Kill to achieve that pattern ? WDYT ? Perhaps we don't need to incur the complexity of this as of now, but just wanted to get an idea of what the direction would be for context/cancellation in client<->server for csi-proxy in general ?

client/api/filesystem/v1alpha1/api.proto
ddebroy: We will not allow arbitrary absolute paths as was [discussed](https://github.com/kubernetes/enhancements/pull/1184#discussion_r308935813) during KEP review. May be I will add a note here clarifying that.

internal/os/filesystem/api.go
ddebroy: Yes we should definitely use the ctx for more the more complex/time consuming operations like formatting a partition, etc as part of the disk APIs. The ctx is passed through to the server layer and ignored for now as discussed [above](https://github.com/kubernetes-csi/csi-proxy/pull/11#discussion_r339925937)

internal/server/filesystem/server.go
ddebroy: Good point. I was thinking about this along the lines of mkdir/rmdir APIs in golang where the absolute or relative nature of a path is determined by the API implementation rather than having to be explicitly specified by a parameter from the caller. Since the nature of the path is trivial to determine (especially given our scoped support surface i.e. no UNC paths, etc), I was thinking we can do without an extra parameter.

I think under most scenarios CSI node plugins typically deal with absolute paths passed by the in-tree CSI plugin in kubelet as required by the spec (https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume and https://github.com/container-storage-interface/spec/blob/master/spec.md#nodestagevolume, etc).

Do you have any specific scenario in mind where there might be ambiguity?

internal/server/filesystem/server.go
ddebroy: I don't see a `pathlen` builtin. Can you point me to the golang package that contains `pathlen` please @kkmsft

kkmsft: Sorry! for the confusion. I was not talking about a package, but the variable being used in the codepath - pathlen  - instead of calculating len(path)

kkmsft:
```suggestion
	if pathlen > 0 && (path[0] == '\\') {
```

internal/server/filesystem/server.go
ddebroy: Decision is to allow Absolute paths only for now. Relative path support can be added later if it becomes necessary.

ddebroy:
@jingxu97 @wk8 @kkmsft this is ready for another look.

Tests:
```
/go/src/github.com/kubernetes-csi/csi-proxy # GO111MODULE=on GOOS=windows go build -o bin/csi-proxy.exe ./cmd/server
/go/src/github.com/kubernetes-csi/csi-proxy # GO111MODULE=on go test github.com/kubernetes-csi/csi-proxy/internal/server/filesystem
ok  	github.com/kubernetes-csi/csi-proxy/internal/server/filesystem	0.009s
```
Windows:
```
PS C:\Users\docker\csi-proxy-2\csi-proxy-1\integrationtests> go test filesystem_test.go
ok      command-line-arguments  0.348s
```

internal/server/filesystem/server.go
jingxu97: is the code only allow absolute path?

cmd/server/main.go
```diff
@@ -1,20 +1,37 @@
...
 }
 
 // apiGroups returns the list of enabled API groups.
-func apiGroups() []server.APIGroup {
-	return []server.APIGroup{
-		&filesystem.Server{},
+func apiGroups() ([]srvtypes.APIGroup, error) {
+	fssrv, err := filesystemsrv.NewServer(*kubeletCSIPluginsPath, *kubeletPodPath, filesystemapi.New())
+	if err != nil {
+		return []srvtypes.APIGroup{}, err
```
wk8: Fine for now, but going forward we might want to have a `Config` struct, and have a standard init func taking a `Config` and returning a `(srvtypes.APIGroup, error)` tuple, so we can daisy chain inits.

More of a note for later, good for now :)

internal/os/filesystem/api.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// Implements the Filesystem OS API calls. All code here should be very simple
+// pass-through to the OS APIs. Any logic around the APIs should go in
+// internal/server/filesystem/server.go so that logic can be easily unit-tested
+// without requiring specific OS environments.
+
+type APIImplementor struct{}
+
+func New() APIImplementor {
```
wk8: Nit: `NewImplementor` ?

internal/server/filesystem/server.go
```diff
@@ -2,24 +2,203 @@ package filesystem
...
+	// for Windows check for C:\\.. prefix only
+	// UNC prefixes of the form \\ are not considered
+	// absolute in the context of CSI proxy
+	if absPathRegexWindows.MatchString(path) {
+		return true
+	}
+	return false
+}
+
+func (s *Server) validatePathWindows(pathCtx internal.PathContext, path string) error {
```
jingxu97: could we move these functions into util so that other apigroup can use it?

PR #14, opened 2019-10-30
Enable building with 1.13 (besides 1.12)

wk8:
**What type of PR is this?**
/kind feature

Upgrading to go 1.13

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ compile: compile-client compile-server compile-csi-proxy-api-gen
 
 .PHONY: compile-client
 compile-client:
-	$(GO_ENV_VARS) go build ./client/...
+	cd client && $(GO_ENV_VARS) go build ./...
 
 .PHONY: compile-server
 compile-server:
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ replace (
 	// https://github.com/kubernetes/gengo/pull/155#issuecomment-537589085
 	// is implemented, and the generic conversion generator merged into code-generator
 	// FIXME: switch back to the upstream repo and/or code-generator!
-	k8s.io/gengo => github.com/wk8/gengo v0.0.0-20191001015530-3d2530bfe606ffd99a90d70ef781861042e23a6f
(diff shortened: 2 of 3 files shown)


msau42:
Is there any particular feature of go1.13 you need? Right now the go version is controller by release-tools/prow.sh, which we need to apply to all the k8s sidecars. cc @pohly

wk8:
> Is there any particular feature of go1.13 you need? Right now the go version is controller by release-tools/prow.sh, which we need to apply to all the k8s sidecars. cc @pohly

@msau42 : None. This is mostly at KK's request, for ease of development.

That being said, this patch only accounts for some minor changes that 1.13 needs. Still runs just fine with 1.12.

ddebroy:
/approve

May be the title of the PR can be changed to "enable building with 1.13 (besides 1.12)".

wk8:
> May be the title of the PR can be changed to "enable building with 1.13 (besides 1.12)".

Great suggestion :)

Makefile
```diff
@@ -22,7 +22,7 @@ compile: compile-client compile-server compile-csi-proxy-api-gen
 
 .PHONY: compile-client
 compile-client:
-	$(GO_ENV_VARS) go build ./client/...
+	cd client && $(GO_ENV_VARS) go build ./...
```
msau42: Sorry this is a little unrelated but I missed the recent changes to the Makefile. Do we want to test compilation and run unit tests as part of the pull job? If so, we need a "make test" rule here.  By moving away from release-tools/build.make, we also miss some rules for verifying gofmt and vendor too

ddebroy: Agree. @wk8 do you think the Makefile was csi-proxy can be reconciled with the make targets under https://github.com/kubernetes-csi/csi-proxy/blob/master/release-tools/build.make? I.e. the common items can be driven by build.make while the special requirements like protobuf gen, code gen, etc. can be driven by the top level Makefile.

I think ^ should be handled in a separate PR.

ddebroy:
/lgtm

The makefile re-alignment will happen in a separate PR from @wk8

PR #15, opened 2019-11-18
Disk APIs

ddebroy:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Introduces the disk device APIs

Fixes #

WIP at the moment

```release-note
NONE
```


internal/server/disk/server.go
```diff
@@ -0,0 +1,78 @@
+package disk
+
+import (
+	"context"
+
+	"github.com/kubernetes-csi/csi-proxy/client/apiversion"
+	"github.com/kubernetes-csi/csi-proxy/internal/server/disk/internal"
+	log "github.com/sirupsen/logrus"
```
kkmsft: Do we need to use logrus ? Can't we do with klog ? Although the request tracing is better in logrus there are only few levels of verbosity. Also, since k8s is in the klog, wouldn't it make sense to align ?

internal/os/disk/api.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (APIImplementor) IsDiskInitialized(diskID string) (bool, error) {
+	cmd := fmt.Sprintf("Get-Disk -Number %s | Where partitionstyle -eq 'raw'", diskID)
+	out, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("error checking initialized status of disk %s: %v, %v", diskID, out, err)
+	}
+	if len(out) == 0 {
+		// disks with raw initializtion not detected
```
kkmsft: nit: initialization

internal/os/disk/api.go
```diff
@@ -0,0 +1,61 @@
...
+// without requiring specific OS environments.
+
+type APIImplementor struct{}
+
+func New() APIImplementor {
+	return APIImplementor{}
+}
+
+func (APIImplementor) IsDiskInitialized(diskID string) (bool, error) {
+	cmd := fmt.Sprintf("Get-Disk -Number %s | Where partitionstyle -eq 'raw'", diskID)
```
kkmsft: Shouldn't we be looking for partition style MBR since in the initialization we are using MBR as the partition style ?

internal/os/disk/api.go
```diff
@@ -0,0 +1,61 @@
...
+	cmd := fmt.Sprintf("Initialize-Disk -Number %s -PartitionStyle MBR", diskID)
+	out, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error initializing disk %s: %v, %v", diskID, out, err)
+	}
+	return nil
+}
+
+func (APIImplementor) PartitionsExist(diskID string) (bool, error) {
+	cmd := fmt.Sprintf("Get-Partition | Where DiskNumber -eq %s", diskID)
```
kkmsft: Shouldn't we be doing Get-disk -DevidId <disk num> | get-partition so that we iterate through only eat the disk we need ?

internal/server/disk/internal/types.go
```diff
@@ -0,0 +1,43 @@
...
+	Target  string
+	LUNID   string
+}
+
+type ListDiskIDsRequest struct {
+}
+
+type ListDiskIDsResponse struct {
+	// Map of disk device IDs and SCSI IDs associated with each disk device
+	Disk_IDs map[string]*DiskIDList
```
kkmsft: Is there a reason why we have to use _ ? Would it be possible to keep it consistent with the camelcase names ?

internal/server/disk/internal/types.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+type ListDiskLocationsResponse struct {
+	// Map of disk device IDs and <adapter, bus, target, lun ID> associated with each disk device
+	DiskLocations map[string]*DiskLocation
+}
+
+type PartitionDiskRequest struct {
+	// Disk device ID of the disk to partition
+	DiskId string
```
kkmsft: nit: DiskId ==> DiskID ?

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/kubernetes-csi/csi-proxy/client v0.0.0-00010101000000-000000000000
 	github.com/pkg/errors v0.8.1
 	github.com/sergi/go-diff v1.0.0
+	github.com/sirupsen/logrus v1.4.1
```
kkmsft: Could we start using klog ?

internal/os/disk/api.go
ddebroy: We want to look for `raw` here. `MBR`/`GPT` would indicate disk is already initialized.

internal/os/disk/api.go
ddebroy: The `DiskNumber -eq diskID` should only return the disk device we are interested in.

internal/os/disk/api.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	if len(out) > 0 {
+		// disk has paritions in it
+		return true, nil
+	}
+	return false, nil
+}
+
+func (APIImplementor) CreatePartition(diskID string) error {
+	cmd := fmt.Sprintf("New-Partition -DiskNumber %s -UseMaximumSize", diskID)
```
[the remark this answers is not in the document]
ddebroy: We do not want to use `DriveLetters` in csi-proxy as much as possible.

client/api/disk/v1alpha1/api.proto
```diff
@@ -0,0 +1,49 @@
...
+    // disk devices enumerated by the host
+    rpc ListDiskLocations(ListDiskLocationsRequest) returns (ListDiskLocationsResponse) {}
+
+    // PartitionDisk initializes and partitions a disk device (if the disk has not
+    // been partitioned already) and returns the resulting volume device ID
+    rpc PartitionDisk(PartitionDiskRequest) returns (PartitionDiskResponse) {}
+
+    // Rescan refreshes the host's storage cache
+    rpc Rescan(RescanRequest) returns (RescanResponse) {}
+}
```
jingxu97: do we also need format and initilize_disk apis?

client/api/disk/v1alpha1/api.proto
```diff
@@ -0,0 +1,49 @@
...
+    map <string, DiskLocation> disk_locations = 1;
+}
+
+message PartitionDiskRequest {
+    // Disk device ID of the disk to partition
+    string diskID = 1;
+}
+
+message PartitionDiskResponse {
+    // Intentionally empty
```
jingxu97: we don't need any result from partition?

client/api/disk/v1alpha1/api.proto
ddebroy: As discussed during the meeting today, `format` is taken care of as part of the volume APIs - https://github.com/kubernetes-csi/csi-proxy/blob/master/client/api/volume/v1alpha1/api.proto#L16 and disk initialization is part of Disk Partitioning: https://github.com/kubernetes-csi/csi-proxy/pull/15/files#diff-87dc7f1dc074717f7df0d9698f32ce1fR51

client/api/disk/v1alpha1/api.proto
ddebroy: For errors, the request can directly return it (e.g. at https://github.com/kubernetes-csi/csi-proxy/pull/15/files#diff-87dc7f1dc074717f7df0d9698f32ce1fR79).

To obtain the Windows volume objects (resulting from partition creation on a disk), a plugin is expected to use `ListVolumesOnDisk`: https://github.com/kubernetes-csi/csi-proxy/blob/master/client/api/volume/v1alpha1/api.proto#L8

PR #17, opened 2019-12-06
Volume APIs

kkmsft:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Add volume APIs

Includes the following as well:
- Fixes the auto generators to reflect the move of types.go
- Freshly generated filesystem auto-gen files

Fixes #

WIP

```release-note
NONE
```


jingxu97:
@kkmsft 

I am wondering whether it is cleaner to have separate PRs. One is for adding new Volume APIs, and the other is to fix existing filesystem APIs? Thanks!

kkmsft:
Looks like need to add the following:
1. Pass format filesystem type and options in FormatVolume.
2. Pass mount options during mounting.

integrationtests/volume_test.go
```diff
@@ -0,0 +1,305 @@
...
+		t.Fatal("Volume formatted. Unexpected !!")
+	}
+
+	formatVolumeRequest := &v1alpha1.FormatVolumeRequest{
+		VolumeId: volumeID,
+	}
+	_, err = client.FormatVolume(context.TODO(), formatVolumeRequest)
+	if err != nil {
+		t.Fatalf("Volume format failed. Error: %v", err)
+	}
```
ddebroy: Will it be possible to have the verification of `FormatVolume` be driven through Powershell directly rather than through `IsVolumeFormatted`? I understand the volume APIs are very thin wrappers around the powershell calls but that would be better since you are not using the API/interface you are testing to perform the validation as well. Sort of along the lines of https://github.com/kubernetes-csi/csi-proxy/blob/master/integrationtests/filesystem_test.go#L49 where a direct OS call is made rather than to the FS API's PathExists API.

integrationtests/volume_test.go
```diff
@@ -0,0 +1,305 @@
...
+	if !isVolumeFormattedResponse.Formatted {
+		t.Fatal("Volume should be formatted. Unexpected !!")
+	}
+
+	resizeVolumeRequest := &v1alpha1.ResizeVolumeRequest{
+		VolumeId: volumeID,
+		// Resize from 5G to 2G
+		Size: 2 * 1024 * 1024 * 1024,
+	}
+
```
ddebroy: Will it be possible to validate the resized volume's size to appropriate size did happen through powershell here?

integrationtests/volume_test.go
```diff
@@ -0,0 +1,305 @@
...
+	// Mount the volume
+	mountVolumeRequest := &v1alpha1.MountVolumeRequest{
+		VolumeId: volumeID,
+		Path:     mountPath,
+	}
+	_, err = client.MountVolume(context.TODO(), mountVolumeRequest)
+	if err != nil {
+		t.Fatalf("Volume id %s mount to path %s failed. Error: %v", volumeID, mountPath, err)
+	}
```
ddebroy: Please add a check/validation here through powershell/OS calls.

client/api/volume/v1alpha1/api.proto
```diff
@@ -0,0 +1,69 @@
...
+    rpc FormatVolume(FormatVolumeRequest) returns (FormatVolumeResponse) {}
+    // ResizeVolume performs resizing of the partition and file system for a block based volume
+    rpc ResizeVolume(ResizeVolumeRequest) returns (ResizeVolumeResponse) {}
+}
+message ListVolumesOnDiskRequest {
+    // Disk device ID of the disk to query for volumes
+    string disk_id = 1;
+}
+message ListVolumesOnDiskResponse {
+    // Volume device IDs of volumes on the specified disk
```
ddebroy: Can you clarify the `\\.\Volume{GUID} format` ID format in the comments wherever volume_id[s] are involved?

internal/os/volume/api.go
```diff
@@ -0,0 +1,94 @@
...
+	cmd := fmt.Sprintf("Get-Volume -UniqueId \"%s\" | Get-Partition | Remove-PartitionAccessPath -AccessPath %s", volumeID, path)
+	out, err := runExec(cmd)
+	if err != nil {
+		return fmt.Errorf("error getting driver letter to mount volume %s: %v, %v", volumeID, out, err)
+	}
+	return nil
+}
+
+// ResizeVolume - resize the volume to the size specified as parameter.
+func (VolAPIImplementor) ResizeVolume(volumeID string, size int64) error {
```
andyzhangx: we can only expand volume, right? shall we check the original size, make sure it's bigger or just depends on the `Resize-Partition` command return info?

internal/server/volume/server.go
```diff
@@ -0,0 +1,144 @@
...
+	hostAPI API
+}
+
+type API interface {
+	ListVolumesOnDisk(diskID string) (volumeIDs []string, err error)
+	// MountVolume mounts the volume at the requested global staging path
+	MountVolume(volumeID, path string) error
+	// DismountVolume gracefully dismounts a volume
+	DismountVolume(volumeID, path string) error
+	// IsVolumeFormatted checks if a volume is formatted with NTFS
```
andyzhangx: is there a fstype parameter somewhere, if it's empty, then it's NTFS

internal/server/volume/server.go
```diff
@@ -0,0 +1,144 @@
...
+type Server struct {
+	hostAPI API
+}
+
+type API interface {
+	ListVolumesOnDisk(diskID string) (volumeIDs []string, err error)
+	// MountVolume mounts the volume at the requested global staging path
+	MountVolume(volumeID, path string) error
+	// DismountVolume gracefully dismounts a volume
+	DismountVolume(volumeID, path string) error
```
andyzhangx: `UnmountVolume` is better

client/api/volume/v1alpha1/api.proto
```diff
@@ -0,0 +1,69 @@
+syntax = "proto3";
+
+package v1alpha1;
+
+service Volume {
+    // ListVolumesOnDisk returns the volume IDs (in \\.\Volume{GUID} format) for
+    // all volumes on a Disk device
+    rpc ListVolumesOnDisk(ListVolumesOnDiskRequest) returns (ListVolumesOnDiskResponse) {}
```
andyzhangx: what do you mean `all volumes on a Disk device`? volume means `partition` of a disk on windows?

kkmsft: In the context here, I think volume can be considered equal to partition which is already formatted. @ddebroy - thoughts ?

integrationtests/volume_test.go
kkmsft: Sure, will change to validate using regular powershell call since it makes sense to do validations using standard mechanism.

integrationtests/volume_test.go
kkmsft: Will add.

internal/server/volume/server.go
kkmsft: Yes, a few more parameters have to be added. Will do a follow up PR with that:
https://github.com/kubernetes-csi/csi-proxy/pull/17#issuecomment-571007850

internal/server/volume/server.go
kkmsft: Although I agree with you on this, Dismount is the terminology used in powershell world. Hence keeping it as Dismount for similarity in this context.

internal/os/volume/api.go
kkmsft: In some conditions it can also reduce the size. For example if there is no data yet the volume size can be reduced as done in the integrated tests in this PR. 

@andyzhangx, @ddebroy  - What do we need to support from CSI driver's perspective - do we need to support reduction as well  ?

We can query the volume, calculate the expected size and give out the error if they are not as expected, What do you say - @andyzhangx ?

client/api/volume/v1alpha1/api.proto
kkmsft: Sure, will do.

client/api/volume/v1alpha1/api.proto
kkmsft: If we are formatting the partitions with -UseMaximumSize we will have only a single volume present. If we have to start supporting multiple partitions then we need to start looking at how to specify those format requirements at the beginning of the workflow.

kkmsft: /cc @jingxu97

kkmsft: get-disk | get-partition| get-volume is what works :-)

internal/os/volume/api.go
jingxu97: I don't think we support reduce the disk size

internal/server/volume/server.go
ddebroy: I am fine with either and we discussed this briefly. To @andyzhangx 's point, `unmount` is used everywhere in k8s codebase.

PR #19, opened 2019-12-18
Add SMB APIs to csi-proxy

jingxu97:
(NOT READY FOR REVIEW) This PR is to add SMB APIs to csi-proxy

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


kkmsft:
@jingxu97 - looks like binary files have been included into the PR. Please remove them.

integrationtests/utils.go
```diff
@@ -12,13 +12,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/kubernetes-csi/csi-proxy/internal/server"
+	srvtypes "github.com/kubernetes-csi/csi-proxy/internal/server/types"
 	"github.com/pkg/errors"
 	"github.com/sergi/go-diff/diffmatchpatch"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-
```
kkmsft: Is this change from gofmt or before gofmt was run ?

internal/os/smb/api.go
```diff
@@ -0,0 +1,80 @@
...
+	if output, err := exec.Command("cmd", "/c", "mklink", "/D", localPath, remotePath).CombinedOutput(); err != nil {
+		klog.Errorf("mklink failed: %v, source(%q) target(%q) output: %q", err, remotePath, localPath, string(output))
+		return fmt.Errorf("MountSmbShare failed: %v, output: %q", err, string(output))
+	}
+	return nil
+}
+
+func (APIImplementor) UnmountSmbShare(remotePath, localPath string) error {
+	klog.V(4).Infof("Unmount target (%q)", localPath)
+	target := normalizeWindowsPath(localPath)
```
kkmsft: If we are using the smb mount mutex during the Mount, shouldn't we use it here to ensure there are no mount/unmount operation going on during the same time ?

internal/os/smb/api.go
```diff
@@ -0,0 +1,80 @@
...
+	if output, err := exec.Command("cmd", "/c", "rmdir", target).CombinedOutput(); err != nil {
```
kkmsft: Wondering whether the SMB mapping should be removed first and then the directory should be removed ? Is there a reason why this order is followed ?

jingxu97:
cc @andyzhangx

internal/os/smb/api.go
jingxu97: actually we probably should not remove smb mapping due to
1. In current csi implementation, the source of the smb mapping will not be available during NodeUnpublishVolume call
2. In case the mapping is shared among different pods, we cannot remove it.

andyzhangx: correct, current smb mapping is GlobalSMBMapping which means mapping could be shared between different pods

internal/os/smb/api.go
andyzhangx: @kkmsft  you mean there is no mount/unmount operation by other component at the same time? smb mount mutex is only to make sure there are no concurrent mount/unmount operation on this proxy.

kkmsft:
@jingxu97 - Can we please close this, since the active PR is - https://github.com/kubernetes-csi/csi-proxy/pull/23

jingxu97:
open #23 and close this one.

PR #21, opened 2020-01-17
IsMountPoint implementation

kkmsft:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

IsLikelyNotMountPath implementation.

Fixes #

```release-note
NONE
```


kkmsft:
@ddebroy - Added integration test.

client/api/filesystem/v1alpha1/api.proto
```diff
@@ -16,6 +16,9 @@ service Filesystem {
     // LinkPath creates a local directory symbolic link between a source path 
     // and target path in the host's filesystem
     rpc LinkPath(LinkPathRequest) returns (LinkPathResponse) {}
+
+    //IsLikelyNotMountPoint checks if a given path is mount or not
```
msau42: Does Windows have the same problem as linux where we a bind mount on the rootfs can't be detected as a mount point?

This is the reason why in Linux we have IsLikelyNotMountPoint and IsNotMountPoint.

If we don't have the same problem in Windows, then I would just go with an "IsNotMountPoint" call, or maybe even "IsMountPoint" so we don't have a negative in the API

kkmsft:
@ddebroy @jingxu97: Updated the PR to move the calls to IsMountPoint. PTAL.

client/api/filesystem/v1alpha1/api.proto
kkmsft: Thank you @msau42. Moved the calls to IsMountPoint.

PR #22, opened 2020-01-18
Enable unit test

jingxu97:
enable unit test by uncommenting main in .prow.sh

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,9 @@ BUILD_TOOLS_DIR = $(BUILD_DIR)/tools
 
 GO_ENV_VARS = GO111MODULE=on GOOS=windows
 
+# TODO: temporarily disable generate and lint because they are not working.
 .PHONY: all
-all: generate compile lint
+all: compile test
 
 .PHONY: compile
 compile: compile-client compile-server compile-csi-proxy-api-gen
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -74,6 +74,14 @@ lint: $(GOLANGCI_LINT)
 	$(GO_ENV_VARS) $(GOLANGCI_LINT) run
 	git --no-pager diff --exit-code
 
+.PHONY: test-go
+test: test-go
+test-go:
+	@ echo; echo "### $@:"
+	# TODO: After issue https://github.com/microsoft/go-winio/pull/169 is resolved, remove the filter on the test path.
```
kkmsft: I was thinking we could have the target set as the following which is more inclusive at the same time worksarounds the issues with go-winio:

`@go test -race -count=1 $(shell go list ./... | grep -v integrationtests | grep -v server$$)`


just not including the integration tests and files at the top level of server directory (that seems to be causing the issue with winio) but includes everthing else and other directories in server directory…

Also the -race to check for races and -count to ensure that cache is not used ?

ddebroy:
Do we have an exception agreed for csi-proxy to not use the prow.sh from https://github.com/kubernetes-csi/csi-release-tools?

client/api/volume/v1alpha1/api.pb.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
-// source: api.proto
```
kkmsft: @jingxu97 - I had raised a PR for that, which is already merged. Can you please rebase this PR with master ?

kkmsft:
> Do we have an exception agreed for csi-proxy to not use the prow.sh from https://github.com/kubernetes-csi/csi-release-tools?

@msau42 @jingxu97 : Since @ddebroy  is out of office could not clarify with him about yet, But would you have some guidance on his question above ^^ ?

Makefile
jingxu97: it is still giving error when run go list 
"build golang.org/x/sys/windows: cannot load golang.org/x/sys/windows: no Go source files"

kkmsft:
@jingxu97 - The PR needs rebase and Release note to be added. 
Otherwise
/lgtm
/approve

PR #30, opened 2020-03-06
Fix smb test

jingxu97:
This PR enable unit tests so that they can be triggered during
presubmit. It also fixes an issue in smb unit test.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/integrationtests/smb_test.go
+++ b/integrationtests/smb_test.go
@@ -7,7 +7,7 @@ import (
 	"math/rand"
 	"os"
 	"os/exec"
-        "strings"
+    "strings"
 	"time"
 
 	"testing"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -15,7 +15,7 @@ BUILD_TOOLS_DIR = $(BUILD_DIR)/tools
 GO_ENV_VARS = GO111MODULE=on GOOS=windows
 
 .PHONY: all
-all: generate compile lint
```
kkmsft: Is there a reason why we don't want to perform generate ? Wouldn't that result in folks checking in code without running generate ?

.prow.sh
```diff
@@ -9,4 +9,4 @@
 
 # TODO: Once we add windows build rules and a Makefile, we can enable
```
kkmsft: Is this TODO comment still required ?

Makefile
jingxu97: it does not work without installing protoc etc.  I am not sure whether we can install it easily.

.prow.sh
jingxu97: no, will remove it.

Makefile
kkmsft: I see. Is there a way for us to run specific make targets ?
If so we could run :
```
make compile
make test
```

and get a compile and run the test ?

integrationtests/smb_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"math/rand"
 	"os"
 	"os/exec"
-        "strings"
```
kkmsft: Since there are many test fixes in this, could we do it as separate PRs, so that we can merge one and work on the comments for another one ?

jingxu97: I spitted it
https://github.com/kubernetes-csi/csi-proxy/pull/22

PR #33, opened 2020-03-23
Build info display, go fmt and go tidy

kkmsft:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Build info display, go fmt and go tidy

Fixes #

```release-note
NONE
```


Makefile
```diff
@@ -12,6 +12,10 @@ REPO_ROOT = $(CURDIR)
 BUILD_DIR = build
 BUILD_TOOLS_DIR = $(BUILD_DIR)/tools
 
+GIT_COMMIT :=$$(git rev-parse --short HEAD)
+SERVER_VERSION ?=v1.0-alpha
```
msau42: It would be better to auto generate this version string from git tags. Otherwise, every release, you need to remember to update this field.

Also it would be better to use 0.1.0 to be consistent with semantic versioning of the other repos.

Also, instead of having this make file be more and more custom for this repo, can we considate these build rules into release-tools/build.make? I assume we're going to need similar things for the node-driver-registrar?

kkmsft: @msau42 : Thank you for the comments. 

I will do the git tag based version. One quick question - Is there any prefix or suffix used to indicate "alpha" or 0.1.0 is enough when we tag the alpha version ? 

Not sure, we can get the build rules merge with release-tools/build.make done before releasing alpha this week. Would it be okay if we take that up after the alpha release. Have raised #34 to keep track of this.

msau42: No "alpha" suffix is needed. In the main README and release notes, just call out that the Status of this project is alpha.

Does the build-rule defined [here](https://github.com/kubernetes-csi/csi-proxy/blob/9118629e2530fcfefe6bce92e8575107b93c85f5/release-tools/build.make#L66) not work for windows? At least from first glance all the arguments seem the same.

kkmsft: On windows Make does not come by default. So we need to use docker to build the binaries as well. In csi proxy we build the windows binaries from linux/macOS, typically for development. I don't think we can use the makefile eco system currently to build out the windows binary from an automated build standpoint of view. Please find more details of this in the PR raised by Chauncey on node-driver-registrar: https://github.com/kubernetes-csi/node-driver-registrar/pull/66. Summary is that current release tools does not work for windows. We need to discuss and arrive a consensus on what approach we want to take here and hence don't think its possible in short term to release tools. Thoughts ?

/cc @ddebroy @jingxu97

msau42: I'm confused about Windows not supporting make because this file itself is a Makefile? I thought the issue was not that make isn't supported, but it was that the build rule was invoking bash? Can we just reuse this build-% rule from the common makefile to make sure that:
- we're using the same version of golang
- "go build" is invoked with the exact same arguments
- the version strings and git tags follow the same format

jingxu97: I opened PR https://github.com/kubernetes-csi/csi-proxy/pull/38 to address this.

kkmsft:
Closing this since #41 will take care of the version printing.

PR #35, opened 2020-03-24
Smb changes to add linking to local path and Stat fix

kkmsft:
**What type of PR is this?**
>

/kind api-change

> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

#### Smb changes to add linking to local path
STAGE phase in CSI Driver would need to ensure that smb mapping is done as well as a link to the SMB share created locally. This PR introduced the local path parameter to the SMB API (optional). If local path parameter is provided, we first check if the path is within the plugin context and then use New-Item powershell command to create a directory softlink. Note that using os.Symlink was giving us errors with container runtime (docker) when starting the container. 

We reuse the the path validation from file server structure by including file server as an input parameter for SMB server initialization.

#### Lstat fix
Currently we use os.Stat to check the presence of a file. If we are looking at a softlink, os.Stat gives access denied. This PR also includes a fix to use Lstat instead of Stat so that it works with soft links. This is to support a scenario is file share driver cases where you have two level links: SMB->Stage Path -> Publish Path. 

Fixes #

```release-note
NONE
```


internal/os/smb/api.go
```diff
@@ -14,8 +15,40 @@ func New() APIImplementor {
...
+	cmd := exec.Command("powershell", "/c", cmdLine)
+	cmd.Env = append(os.Environ(),
+		fmt.Sprintf("smbremotepath=%s", remotePath))
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("error checking smb mapping %s, err: %v", remotePath, err)
+	}
+
+	if len(out) == 0 || !strings.EqualFold(strings.TrimSpace(string(out)), "OK") {
```
jingxu97: if there is no output, does it mean it is not mapped?

internal/os/smb/api.go
```diff
@@ -14,8 +15,40 @@ func New() APIImplementor {
...
+		return false, nil
+	}
+	return true, nil
+}
+
+func (APIImplementor) SMBLink(remotePath, localPath string) error {
```
jingxu97: this is a generic way to create symlink, do we want to only put it in smb api?

internal/server/filesystem/server.go
```diff
@@ -67,6 +67,10 @@ func isAbsWindows(path string) bool {
 	return absPathRegexWindows.MatchString(path)
 }
 
+func (s *Server) ValidateSMBLinkPath(path string) error {
```
jingxu97: can we just use validataPathWindows directly?

internal/server/smb/server.go
```diff
@@ -5,36 +5,63 @@ import (
...
+
+	if !isMapped {
+		klog.V(4).Infof("Remote %s not mapped. Mapping now!", remotePath)
+		err := s.hostAPI.NewSmbGlobalMapping(remotePath, localPath, request.Username, request.Password)
+		if err != nil {
+			return response, err
+		}
 	}
+
+	if len(localPath) != 0 {
```
ddebroy: Is the `localPath` optional?

internal/os/smb/api.go
ddebroy: Please put a comment here explaining some of the subtle aspects around links that you found out when looking into this and why we need a special mechanism for SMB.

internal/server/filesystem/server.go
kkmsft: I tried but could not workout the export due to the need to use the Using ValidatePathWindows requires us to export the type - internal.PLUGIN. Due to restrictions around the way internal can be used in go, this was not working when used from smb - would not compile. Hence had to opt to use in an interface instead of using the common code directly.

internal/server/smb/server.go
kkmsft: yes.

internal/server/filesystem/server.go
jingxu97: validatePluginPath as discussed?

internal/os/smb/api.go
kkmsft: Yes, that is what is going to be the best path in that case.

kkmsft:
Have addressed some of the review comments. 
@ddebroy @jingxu97 - PTAL

internal/os/smb/api.go
```diff
@@ -14,8 +15,47 @@ func New() APIImplementor {
...
+	_, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error linking %s to %s, err: %v", remotePath, localPath, err)
+	}
+
+	return nil
+}
+
+func (APIImplementor) NewSmbGlobalMapping(remotePath, localPath, username, password string) error {
+	klog.V(4).Infof("NewSmbGlobalMapping: remotePath:%q, localPath:%q", remotePath, localPath)
```
jingxu97: is there any restrictions on localpath?
Could you also comment on the use of localpath?

internal/server/smb/server.go
```diff
@@ -5,36 +5,63 @@ import (
...
 type Server struct {
-	hostAPI API
+	hostAPI  API
+	fsServer *fsserver.Server
 }
 
 type API interface {
-	NewSmbGlobalMapping(remotePath, username, password string) error
+	IsSmbMapped(remotePath string) (bool, error)
+	SMBLink(remotePath, localPath string) error
```
jingxu97: maybe a API name more with more details? Like CreateRemoteShareLink?

internal/server/smb/server.go
```diff
@@ -5,36 +5,63 @@ import (
...
+	NewSmbGlobalMapping(remotePath, localPath, username, password string) error
```
jingxu97: the localPath in here and the localPath in above SMBLink is different, right?

kkmsft: They are not different. Having said that we don't need the local path in the host APIs since its really not used any more since the link part is moved out. Will update the PR.

internal/server/smb/server.go
kkmsft: To make is similar to the other APIs - NewSmbLlink

internal/server/smb/server.go
```diff
@@ -5,36 +5,63 @@ import (
...
 )
 
 type Server struct {
-	hostAPI API
+	hostAPI  API
+	fsServer *fsserver.Server
 }
 
 type API interface {
+	IsSmbMapped(remotePath string) (bool, error)
```
jingxu97: here isSmbMapped uses remotePath, is there any case to use localPath to verify?

PR #36, opened 2020-03-25
Update release tool and use it in build 

jingxu97:
This PR first update the release tool from upstream and also update
Makefile to use the build target. The rest of the targets still use the
original ones proposed.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


msau42:
/hold
release-tools subtree has direct modifications, which we don't allow, and potentially can cause difficulties down the line when we try to consolidate the release-tools together. We need to fix https://github.com/kubernetes-csi/csi-proxy/issues/37 first.

release-tools/build.make
```diff
@@ -67,8 +67,11 @@ endif
 # separated by semicolon. An empty variable or empty entry (= just a
 # semicolon) builds for the default platform of the current Go
 # toolchain.
+ifdef os
+BUILD_PLATFORMS=$(os)
```
msau42: BUILD_PLATFORMS should be set in the Makefile instead of here.

release-tools/build.make
```diff
@@ -67,8 +67,11 @@ endif
...
+else
 BUILD_PLATFORMS =
-
+endif
 # This builds each command (= the sub-directories of ./cmd) for the target platform(s)
 # defined by BUILD_PLATFORMS.
 build-%: check-go-version-go
```
msau42: I think it's this "check-go-version-go" build rule that is problematic for Windows. We may need to make changes in csi-release-tools to decouple these steps. cc @pohly

Makefile
```diff
@@ -1,8 +1,8 @@
-## This Makefile is meant to be run on Unix hosts - as such, it only supports the few
-## operations that don't require on a Windows host, mostly code generation and linting.
+os=windows
+CMDS=server
```
msau42: "server" seems too generic of a binary name. Should it be called something more specific like "csi-proxy-server"?

release-tools/build.make
pohly: We can add to `build.make`:

```
CHECK_GO_VERSION ?= check-go-version-go

build-%: $(CHECK_GO_VERSION)
```

Then the top-level Makefile can disable the check with:
```
CHECK_GO_VERSION =
include release-tools/build.make
```

But as discussed elsewhere, why don't we fix the check script so that it also works on Windows? Can we assume that POSIX shell scripts work on Windows?

msau42: Agree, I think making it work in Windows is the preferable approach. I'm not sure how much work it's going to take, so a short term workaround for now is to decouple the rule.

Makefile
jingxu97: I will update the dir name in a separate PR to reduce the changes in this one.

jingxu97:
close this PR in favor of #38

PR #41, opened 2020-03-27
Add logs

jingxu97:
This PR adds logs for csi-proxy.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/internal/os/volume/api.go
+++ b/internal/os/volume/api.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
-
-	"k8s.io/klog"
 )
 
 const formatFilesystem = "ntfs"
@@ -19,9 +17,7 @@ func New() VolAPIImplementor {
 }
 
 func runExec(cmd string) ([]byte, error) {
-	klog.V(5).Infof("Running command: %s", cmd)
 	out, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
-	klog.V(5).Infof("Result: %s. Len: %d. Error: %v.", string(out), len(string(out)), err)
 	return out, err
 }
 
@@ -30,7 +26,7 @@ func (VolAPIImplementor) ListVolumesOnDisk(diskID string) (volumeIDs []string, e
 	cmd := fmt.Sprintf("(Get-Disk -DeviceId %s |Get-Partition | Get-Volume).UniqueId", diskID)
 	out, err := runExec(cmd)
 	if err != nil {
-		return []string{}, err
+		return []string{}, fmt.Errorf("error list volumes on disk. cmd: %s, output: %s, error: %v", cmd, string(out), err)
 	}
 
 	volumeIds := strings.Split(strings.TrimSpace(string(out)), "\r\n")
@@ -42,7 +38,7 @@ func (VolAPIImplementor) FormatVolume(volumeID string) (err error) {
 	cmd := fmt.Sprintf("Get-Volume -UniqueId \"%s\" | Format-Volume  -FileSystem %s -Confirm:$false", volumeID, formatFilesystem)
 	out, err := runExec(cmd)
 	if err != nil {
-		return fmt.Errorf("error formatting volume %s: %v, %v", volumeID, out, err)
+		return fmt.Errorf("error formatting volume. cmd: %s, output: %s, error: %v", cmd, string(out), err)
 	}
 	// TODO: Do we need to handle anything for len(out) == 0
 	return nil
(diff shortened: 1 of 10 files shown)


internal/server/smb/server.go
```diff
@@ -66,6 +67,7 @@ func (s *Server) NewSmbGlobalMapping(context context.Context, request *internal.
 }
 
 func (s *Server) RemoveSmbGlobalMapping(context context.Context, request *internal.RemoveSmbGlobalMappingRequest, version apiversion.Version) (*internal.RemoveSmbGlobalMappingResponse, error) {
+	klog.V(2).Infof("calling RemoveSmbGlobalMapping with remote path %q", request.RemotePath)
```
kkmsft: V(4) - like followed in other cases ?

internal/server/filesystem/server.go
```diff
@@ -83,32 +84,39 @@ func (s *Server) validatePathWindows(pathCtx internal.PathContext, path string)
...
 		prefix = s.kubeletPodPath
 	} else {
+		klog.Errorf("invalid PathContext: %v", pathCtx)
 		return fmt.Errorf("invalid PathContext: %v", pathCtx)
 	}
 
 	pathlen := len(path)
 
 	if pathlen > utils.MaxPathLengthWindows {
+		klog.Errorf("path length %d exceeds maximum characters: %d", pathlen, utils.MaxPathLengthWindows)
```
kkmsft: Logging errors - I would prefer having errors wrapped and we only print this at the top level - in server and can distinguish the error based on wrapped error ? What do you say ?

internal/server/disk/server.go
```diff
@@ -96,13 +97,13 @@ func (s *Server) Rescan(context context.Context, request *internal.RescanRequest
 }
 
 func (s *Server) GetDiskNumberByName(context context.Context, request *internal.GetDiskNumberByNameRequest, version apiversion.Version) (*internal.GetDiskNumberByNameResponse, error) {
+	klog.V(2).Infof("calling GetDiskNumberByName with diskName %q", request.DiskName)
```
kkmsft: V(4) ?

internal/os/volume/api.go
```diff
@@ -19,9 +19,9 @@ func New() VolAPIImplementor {
 }
 
 func runExec(cmd string) ([]byte, error) {
-	klog.V(5).Infof("Running command: %s", cmd)
+	klog.V(4).Infof("Running command: %s", cmd)
```
kkmsft: Would we end up logging any password when logging the command as a whole ? Guessing that only env has the sensitive info like that we wouldn't print it here. Wanted to confirm ?

internal/os/volume/api.go
```diff
@@ -19,9 +19,9 @@ func New() VolAPIImplementor {
...
 	out, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
-	klog.V(5).Infof("Result: %s. Len: %d. Error: %v.", string(out), len(string(out)), err)
+	klog.V(4).Infof("Result: %s. Len: %d. Error: %v.", string(out), len(string(out)), err)
```
kkmsft: May be V(5) here since this is going to be a lot of logs when it comes to powershell so we can do 4 for other server and os layer and 5 for just the powershell. What do you say ?

cmd/server/main.go
```diff
@@ -31,6 +31,8 @@ func main() {
 		panic(err)
 	}
 	s := server.NewServer(apiGroups...)
+
+	klog.Info("starting api server ...")
```
kkmsft: Could you please add the printing of version as well here?

internal/server/filesystem/server.go
jingxu97: one benefit of logging here is we can log the output with error too, I feel sometimes output gives more information, and error only gives some error exit code or status?

cmd/server/main.go
jingxu97: fixed it

internal/os/volume/api.go
jingxu97: I removed the log. Is that ok?

internal/server/filesystem/server.go
jingxu97: i modified. PTAL

internal/server/smb/server.go
jingxu97: fixed it

ddebroy:
Some fmt changes.

internal/server/disk/server.go
jingxu97: fixed

internal/os/volume/api.go
jingxu97: remove the log from os

PR #53, opened 2020-05-21
Add build and release for csi proxy to gcs bucket

jingxu97:
This PR adds building and publishing csi-proxy binary to gcs bucket k8s-artifacts-csi
The prow job for publishing the binary in test-infra is here https://github.com/kubernetes/test-infra/pull/17683

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,32 @@
+# See https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md for more details on image pushing process
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
+# this prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
+# or any new substitutions added in the future.
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  # The image must contain bash and curl. Ideally it should also contain
+  # the desired version of Go (currently defined in release-tools/travis.yml),
+  # but that just speeds up the build and is not required.
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20200421-a2bf5f8'
+    entrypoint: bash
+    env:
+    - GIT_TAG=${_GIT_TAG}
+    - PULL_BASE_REF=${_PULL_BASE_REF}
+    - HOME=/root
+    args:
+    - .cloudbuild.sh
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
+  # _PULL_BASE_REF will contain the ref that was pushed to trigger this build -
+  # a branch like 'master' or 'release-0.2', or a tag like 'v0.2'.
+  _PULL_BASE_REF: 'master'
+artifacts:
+  objects:
+    location: 'gs://k8s-artifacts-csi/csi-proxy/dev'
+    paths: 'bin/csi-proxy-${_PULL_BASE_REF}.exe'
+
--- a/.cloudbuild.sh
+++ b/.cloudbuild.sh
@@ -1,0 +1,9 @@
+#! /bin/bash
+
+. release-tools/prow.sh
+
+# Extract tag-n-hash value from GIT_TAG (form vYYYYMMDD-tag-n-hash) for REV value.
+REV=v$(echo $GIT_TAG | cut -f3- -d 'v')
+
+run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make build REV=${REV}
+cp bin/csi-proxy.exe bin/csi-proxy-${PULL_BASE_REF}.exe


jingxu97:
This PR is ready for review. PTAL. Thanks!

jingxu97:
cc @justaugustus

jingxu97:
cc @verult

build/cloudbuild.yaml
```diff
@@ -0,0 +1,17 @@
+# Build the module.
+steps:
+- name: gcr.io/cloud-builders/go:debian
```
msau42, line 3: Do we want to use the k8s build images that have specific versions of golang, like in https://github.com/kubernetes-csi/csi-release-tools/pull/87/files#diff-e5fc65747fb9417968a3295dc1ac5111R26?

build/cloudbuild.yaml
```diff
@@ -0,0 +1,17 @@
...
+  entrypoint: bash
+  args:
+  - -c
+  - |
+    set -ex
+    IFS=- read str1 rev <<< "${_GIT_TAG}"
+    make build REV=${rev}
+    cp bin/csi-proxy.exe bin/csi-proxy${_GIT_TAG}.exe
```
msau42, line 11: We may want to consider enhancing this later with the logic in https://github.com/kubernetes-csi/csi-release-tools/pull/87/files#diff-841c5d7c9fb7c0ba0fb2f452298ed526R147 to support "canary" images

build/cloudbuild.yaml
```diff
@@ -0,0 +1,17 @@
...
+substitutions:
+  _GIT_TAG: '12345'
+artifacts:
+  objects:
+    location: 'gs://k8s-artifacts-csi'
```
msau42, line 16: This is the "prod" bucket which I believe we should only push to by promoting from staging. Is there a staging bucket we can use instead?

build/cloudbuild.yaml
jingxu97: got it.

build/cloudbuild.yaml
jingxu97: before I confirmed with bartsmykla@ that there is no such process yet.

build/cloudbuild.yaml
jingxu97: I think go:debian get the latest version of go https://github.com/GoogleCloudPlatform/cloud-builders/pull/132. Based on that PR's comment, seems like it also does not use specific version of go. @pohly might help on this?

build/cloudbuild.yaml
justaugustus: Correct, there's no GCS promotion process yet. I talk about it a little here: https://github.com/kubernetes/release/issues/964#issuecomment-564727533

I wrote [`gh2gcs`](https://github.com/kubernetes/release/tree/master/cmd/gh2gcs) very recently, which supports uploading GitHub releases to a GCS bucket.

See the usage here: https://github.com/kubernetes/release/pull/1311

So if you publish release tarballs to GitHub, you'll be able to do something like this to upload to GCS:

```shell
$ time gh2gcs --log-level debug --org kubernetes-csi --repo csi-proxy --bucket k8s-artifacts-csi --release-dir release --tags v0.0.0,v0.0.1,v0.0.2
```

If you run the command w/o the `--tags` flag, it'll iterate over all releases and upload them.
You could add a step in the cloudbuild.yaml to do this.

Eventually, I'll going to update the tool to accept a yaml config, so you can add a config file somewhere, that looks roughly like this:

```yaml
- org: foo
  repo: bar
  tags: []
  gcsBucket: bucket-name
  releaseDir: release-dir
```

Let me know what you think!

build/cloudbuild.yaml
pohly: Please have a look at https://github.com/kubernetes-csi/csi-release-tools/issues/86#issuecomment-634509901 an the PRs linked to there.

In a nutshell, the approach there is to install the right version of the Go toolchain before using it, using the normal build rules for Windows binaries.

pohly: The relevant code is https://github.com/pohly/csi-release-tools/blob/bd416901d4b7cad7bc70a39f508b68e2a9fff1f6/cloudbuild.sh which sources `release-tools/prow.sh` and then invokes image building with `run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make push-multiarch`.

You could do something similar and use `run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make build` to produce the csi-proxy.exe.
```

File: build/cloudbuild.yaml (hunk 3 above)
Comment (jingxu97) [reply build/cloudbuild.yaml]: @justaugustus thank you very much for providing this tool. For now, I will follow other csi repo build step first. Later we will try to see how to improve it with your tool.

File: build/cloudbuild.yaml (hunk 1 above)
Comment (jingxu97) [reply build/cloudbuild.yaml]: got it. Thanks!
I updated it.

**Comment by jingxu97** [comment]
@pohly @msau42 @justaugustus PR is updated. PTAL. thanks!

File: cloudbuild.yaml [hunk 4]
```diff
@@ -0,0 +1,32 @@
...
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
+  # _PULL_BASE_REF will contain the ref that was pushed to trigger this build -
+  # a branch like 'master' or 'release-0.2', or a tag like 'v0.2'.
+  _PULL_BASE_REF: 'master'
+artifacts:
+  objects:
+    location: 'gs://disksize-bucket'
```
msau42, line 30: Is this location correct?

build/cloudbuild.yaml
msau42: I couldn't quite follow the discussion, but is the suggestion to put ci images under a subfolder called "dev/ci", and then "promote" releases to "release/"? That way it's easy for someone to easily distinguish what is a production ready image vs a dev image.

cloudbuild.yaml
jingxu97: oh, sorry, that's for my local testing. fixed it.

jingxu97:
Updated, PTAL @msau42 @pohly

jingxu97:
@pohly @msau42 I updated a small change. PTAL. Thanks!

jingxu97:
@pohly @msau42 I updated with a small change due to naming. Could you please add /lgtm again? Thanks!

cloudbuild.yaml
```diff
@@ -0,0 +1,32 @@
...
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
+  # _PULL_BASE_REF will contain the ref that was pushed to trigger this build -
+  # a branch like 'master' or 'release-0.2', or a tag like 'v0.2'.
+  _PULL_BASE_REF: 'master'
+artifacts:
+  objects:
+    location: 'gs://k8s-artifacts-csi/dev'
+    paths: 'bin/csi-proxy-${_PULL_BASE_REF}.exe'
```
msau42, line 31: Ah sorry one more thought came to me. Do we want to have a "csi-proxy/bin/" directory in case in the future we have more than one binary we want to put here?

jingxu97: you mean the gs bucket location as gs://k8s-artifacts-csi/csi-proxy/bin? Or put it in paths: csi-proxy/bin/csi-proxy-${_PULL_BASE_REF}.exe? 

For the bucket, if there are other binaries, they will have different names? For the paths here, I am not sure adding csi-proxy is useful?

msau42: either way. I think it's better organized when each binary is in separate directories, especially when we start to get many versions stored.

jingxu97: got it. But I thought paths is the local path where the file is stored locally, and location is the remote store where the file will be uploaded. So  we should add it to location?
How about this gs://k8s-artifacts-csi/csi-proxy/dev?

PR #61, opened 2020-07-14
ListDiskIDs API

ksubrmnn:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR introduces the new v1beta1 version of the disk API. It also includes a ListDiskIDs function that currently returns page83 ID

Fixes #

```release-note
New v1beta1 disk API which includes new ListDiskIDs functionality
```


internal/shared/disk/types.go
```diff
@@ -10,3 +10,8 @@ type DiskLocation struct {
 	Target  string
 	LUNID   string
 }
+
+// DiskID definition
+type DiskID struct {
+	Page83 string
```
ddebroy, line 16: See above note about potentially making this more extensible through a map of <ID-type, ID> rather than hardcoding the ID types.

client/api/disk/v1beta1/api.proto
```diff
@@ -0,0 +1,67 @@
...
+message RescanResponse {
+    // Intentionally empty
+}
+
+message ListDiskIDsRequest {
+    // Intentionally empty
+}
+
+message DiskID {
+    string page83 = 1;
```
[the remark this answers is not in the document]
ksubrmnn: I see what you're saying here, and I can make this change.

I'm a bit worried as this wouldn't be strongly typed so users may not know what ID types are available. We could create an enum for ID types but this once again creates an API versioning problem. 

If you think it's not necessary to create strongly typed IDs, then I can implement this change.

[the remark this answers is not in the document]
ddebroy: I would go with loosely typed for now. We can document the types as part of CSI docs. cc @jingxu97 Thanks!

internal/shared/disk/types.go
```diff
@@ -10,3 +10,9 @@ type DiskLocation struct {
 	Target  string
 	LUNID   string
 }
+
+// DiskID definition
+type DiskID struct {
+	// Map of Disk ID types and Disk ID values
+	DiskIDs map[string]string
```
ddebroy, line 17: nit: Maybe you can call this field `Identifiers`?

client/api/disk/v1beta1/api.proto
```diff
@@ -0,0 +1,68 @@
...
+
+message RescanResponse {
+    // Intentionally empty
+}
+
+message ListDiskIDsRequest {
+    // Intentionally empty
+}
+
+message DiskID {
```
ddebroy, line 60: nit: how about calling this `DiskIDs` since the only member is a `map` and therefore will have multiple IDs.

client/api/disk/v1beta1/api.proto
```diff
@@ -0,0 +1,68 @@
...
+    // Map of Disk ID types and Disk ID values
+    map <string, string> diskIDs = 1;
```
ddebroy, line 62: nit: `diskIDs` maybe called `identifiers` to not conflict with `DiskIDs` as suggested above.

PR #66, opened 2020-08-16
fix: SMB remount issue

andyzhangx:
**What type of PR is this?**
/kind bug

Porting PR(https://github.com/kubernetes/utils/pull/184) to csi-proxy.
There is a critical bug in original SMB mount logic:

   When SMB server password changed, there is no way for csi-proxy to remove original SMB server connection, only workaround is user log on to the Windows agent node, and run `Remove-SMBGlobalMapping` command manually to remove that connection. This PR could fix the issue.

Fixes #

/assign @jingxu97 

```release-note
fix: SMB remount issue
```

### DIFF
--- a/internal/os/filesystem/api.go
+++ b/internal/os/filesystem/api.go
@@ -3,6 +3,8 @@ package filesystem
 import (
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 )
 
 // Implements the Filesystem OS API calls. All code here should be very simple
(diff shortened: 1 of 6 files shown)


andyzhangx:
@jingxu97 I think we should make this PR into v0.2.0 release

internal/server/smb/server.go
```diff
@@ -45,6 +45,22 @@ func (s *Server) NewSmbGlobalMapping(context context.Context, request *internal.
 		isMapped = false
 	}
 
+	valid, err := s.fsServer.PathValid(context, remotePath)
```
jingxu97, line 48: Another option is let the caller to perform this logic. csi-proxy just provide all APIs needed such as PathValid.

internal/os/filesystem/api.go
```diff
@@ -31,6 +33,21 @@ func (APIImplementor) PathExists(path string) (bool, error) {
...
+	cmd.Env = append(os.Environ(), fmt.Sprintf("remoteapth=%s", path))
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("returned output: %s, error: %v", string(output), err)
+	}
+
+	return strings.HasPrefix(strings.ToLower(string(output)), "true"), nil
+}
+
+func (APIImplementor) PathValid(path string) (bool, error) {
```
jingxu97, line 47: pls add some comments here.
I think it worth mentioning the password change use case?

andyzhangx: done

andyzhangx:
@jingxu97 PTAL, added a comment and also added a windows unit test for `pathValide` func, and I found there is not windows unit test running for this repo.

internal/server/smb/server.go
andyzhangx: I think we could expose `PathValid` interface in filesystem, and for this issue, let's fix it first.

PR #81, opened 2020-09-28
Add v1alpha1 system API

ksubrmnn:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Introduces v1alpha1 System API for CSI-Proxy
```


client/api/system/v1alpha1/api.pb.go
```diff
@@ -0,0 +1,303 @@
...
+const (
+	// Verify that this generated code is sufficiently up-to-date.
+	_ = protoimpl.EnforceVersion(20 - protoimpl.MinVersion)
+	// Verify that runtime/protoimpl is sufficiently up-to-date.
+	_ = protoimpl.EnforceVersion(protoimpl.MaxVersion - 20)
+)
+
+// This is a compile-time assertion that a sufficiently up-to-date version
+// of the legacy proto package is being used.
+const _ = proto.ProtoPackageIsVersion4
```
ddebroy, line 30: Can you please do `go get -u github.com/golang/protobuf/protoc-gen-go@v1.3.2` to make sure this continues to use the (slightly) older version of `protoc-gen-go` like the rest of the API groups like [this](https://github.com/kubernetes-csi/csi-proxy/blob/master/client/api/disk/v1beta1/api.pb.go#L25)? I would like to understand the consequences of bumping the version of individual API groups on clients a little more before ideally bumping it for all the API groups. If a client plugin is able to handle both v1.4 and v1.3 simultaneously, then I have no concerns.

client/api/system/v1alpha1/api.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package v1alpha1;
+
+option go_package = "github.com/kubernetes-csi/csi-proxy/client/api/system/v1alpha1";
+
+service System {
+    // GetSerialNumber returns the device's serial number
```
ddebroy, line 8: Can you rename it to `GetBIOSSerialNumber` please? I think there will be other system-wide serial numbers we will want to query down the line and having BIOS in the name makes it more specific.

integrationtests/system_test.go
```diff
@@ -0,0 +1,25 @@
...
+		client, err := v1alpha1client.NewClient()
+		require.Nil(t, err)
+		defer client.Close()
+
+		request := &v1alpha1.GetSerialNumberRequest{}
+		response, err := client.GetSerialNumber(context.TODO(), request)
+		require.Nil(t, err)
+		require.NotNil(t, response)
+		require.NotEmpty(t, response.SerialNumber)
+		t.Logf("The serial number is %s", response.SerialNumber)
```
ddebroy, line 23: Can you also invoke the same cmd - `wmic bios get serialnumber` from the test and compare the value? Ideally we could have queried the serial number from other out-of-band mechanism (like through the hypervisor if possible) but that will make the test specific to a certain hypervisor environment.

ddebroy (commented):
Looks good overall. Just a few minor suggestions.

internal/os/system/api.go
```diff
@@ -0,0 +1,39 @@
...
+	lines := strings.FieldsFunc(string(result), func(r rune) bool {
+		switch r {
+		case '\n', '\r':
+			return true
+		default:
+			return false
+		}
+	})
+	if len(lines) != 2 {
+		return "", fmt.Errorf("received unexpected value retrieving vm uuid: %q", string(result))
```
ddebroy, line 36: Can you please remove `vm` from the error string? The host machine where csi-proxy is running need not be a VM.

ddebroy (commented):
/lgtm
/approve

internal/os/system/api.go
ddebroy: Submitted https://github.com/kubernetes-csi/csi-proxy/pull/94 to address ^. @ksubrmnn can you please lgtm it?

PR #84, opened 2020-10-03
Add WriteVolumeCache API

jingxu97:
add writevolumecache api to csi-proxy

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Add WriteVolumeCache API
```


internal/os/volume/api.go
```diff
@@ -47,6 +47,17 @@ func (VolAPIImplementor) FormatVolume(volumeID string) (err error) {
 	return nil
 }
 
+// WriteVolumeCache - Writes the file system cache to disk with given file path
```
andyzhangx, line 50: nit: `Writes the file system cache to disk where given file path is located`

andyzhangx (approved):
so this API is for specific file to flush memory to disk, right?

/lgtm

client/api/volume/v1beta1/api.proto
```diff
@@ -24,6 +24,8 @@ service Volume {
     rpc GetVolumeDiskNumber(VolumeDiskNumberRequest) returns (VolumeDiskNumberResponse) {}
     // GetVolumeIDFromMount gets the volume id for a given mount
     rpc GetVolumeIDFromMount(VolumeIDFromMountRequest) returns (VolumeIDFromMountResponse) {}
+    // WriteVolumeCache write volume cache to disk
+    rpc WriteVolumeCache(WriteVolumeCacheRequest) returns (WriteVolumeCacheResponse) {}
```
msau42, line 28: I think we need a v1beta2 if we're going to modify the API since csi-proxy versions are not tied to k8s versions

jingxu97:
> so this API is for specific file to flush memory to disk, right?
> 
> /lgtm

actually it mainly for volume to cache data before calling unstagevolume. I am considering to use volumeId too. For certain plugins like gce pd, we don't keep volume unique id, instead, it uses mount path. Get-volume could accept either volumeId or file path. But all other volume API currently use uniqueId, so maybe it makes more sense to use volumeId instead of path. 
@ddebroy

client/api/volume/v1beta1/api.proto
jingxu97: updated.

client/api/volume/v1beta2/api.proto
```diff
@@ -0,0 +1,132 @@
...
+    string mount = 1;
+}
+
+message VolumeIDFromMountResponse {
+    // Mount
+    string volume_id = 1;
+}
+
+message WriteVolumeCacheRequest {
+    // Volume device ID of the volume to format
```
ddebroy, line 126: nit: please fix the comment above referring to `format`. Also please use `volume_id` as you mentioned and we discussed.

ddebroy (commented):
One comment to fix a comment and switch to volume_id

client/api/volume/v1beta2/api.proto
jingxu97: updated. PTAL

internal/server/volume/server.go
```diff
@@ -137,7 +139,24 @@ func (s *Server) FormatVolume(context context.Context, request *internal.FormatV
...
 		return response, err
 	}
+	return response, nil
+}
+
+func (s *Server) WriteVolumeCache(context context.Context, request *internal.WriteVolumeCacheRequest, version apiversion.Version) (*internal.WriteVolumeCacheResponse, error) {
+	klog.V(4).Infof("calling WriteVolumeCache with request: %+v", request)
+	response := &internal.WriteVolumeCacheResponse{}
+
+	filePath := request.FilePath
```
ddebroy, line 149: This needs to be. Since the type is `string`, the compiler did not catch it. 
```
	volumeID := request.VolumeId
	if volumeID == "" {
		klog.Errorf("volume id empty")
		return response, fmt.Errorf("volume id empty")
	}
	err := s.hostAPI. WriteVolumeCache(volumeID)
```

ddebroy (commented):
One more nit around volumeID. Otherwise lgtm.

internal/server/volume/server.go
jingxu97: thanks for catching it. updated. PTAL

ddebroy (commented):
/lgtm
/approve

PR #85, opened 2020-10-05
Add path parameter into VolumeStat() API

jingxu97:
Add a second parameter filePath into VolumeStat so that it can use path
to check the volume usage

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


client/api/volume/v1beta1/api.proto
```diff
@@ -91,6 +91,8 @@ message ResizeVolumeResponse {
 message VolumeStatsRequest{
     // Volume device Id of the volume to get the stats for
     string volume_id = 1;
+    // Path where the volume has been mounted.
+    string path = 2;
```
msau42, line 95: I think we should make a v1beta2 if we modify the API in any way. Because csi proxy versions are not tied to K8s versions, it is not possible to tell if the version of csi proxy running on the node supports this new field or not.

jingxu97: agree. updated to v1beta2. PTAL

jingxu97:
@ddebroy @msau42 PTAL. Thanks!

jingxu97:
assign @manueltellez PTAL

client/api/volume/v1beta2/api.proto
```diff
@@ -0,0 +1,123 @@
...
+
+message ResizeVolumeResponse {
+    // Intentionally empty
+}
+
+message VolumeStatsRequest{
+    // Volume device Id of the volume to get the stats for
+    string volume_id = 1;
+    // Path where the volume has been mounted.
+    string path = 2;
```
ddebroy, line 95: The overall convention we have been following is to accept `volume_id` as the only/canonical parameter for the API calls. I think introducing the `path` parameter deviates from this convention. Will it be possible to introduce a `GetVolumeIDFromPath` (like we have `GetVolumeIDFromMount`) and using the response from `GetVolumeIDFromPath` to pass over to `VolumeStats`?

ddebroy (commented):
@jingxu97 I have a suggestion to restructure the API a little bit.

PR #91, opened 2020-10-13
noop when resize is called and size of partition is >= requested size

manueltellez:
**What type of PR is this?**
/kind bug

Avoids an error when the underlying volume size is already the greater than or equal to the requested size. PowerShell in that case returns unsupported size, in reality it should be a noop.

Fixes #

```release-note
NONE
```

### DIFF
--- a/internal/os/volume/api.go
+++ b/internal/os/volume/api.go
@@ -24,6 +24,23 @@ func runExec(cmd string) ([]byte, error) {
 	return out, err
 }
 
+func getVolumeSize(volumeID string) (int64, error) {
+	cmd := fmt.Sprintf("(Get-Volume -UniqueId \"%s\" | Get-partition).Size", volumeID)
+	out, err := runExec(cmd)
+
+	if err != nil || len(out) == 0 {
+		return -1, fmt.Errorf("error getting size of the partition from mount. cmd %s, output: %s, error: %v", cmd, string(out), err)
+	}
+
+	outString := strings.TrimSpace(string(out))
+	volumeSize, err := strconv.ParseInt(outString, 10, 64)
+	if err != nil {
+		return -1, fmt.Errorf("error parsing size of volume %s received %v trimmed to %v err %v", volumeID, out, outString, err)
+	}
+
+	return volumeSize, nil
+}
+
 // ListVolumesOnDisk - returns back list of volumes(volumeIDs) in the disk (requested in diskID).
 func (VolAPIImplementor) ListVolumesOnDisk(diskID string) (volumeIDs []string, err error) {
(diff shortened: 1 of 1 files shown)


internal/os/volume/api.go
```diff
@@ -109,6 +110,27 @@ func (VolAPIImplementor) ResizeVolume(volumeID string, size int64) error {
...
+	err = json.Unmarshal([]byte(outString), &getPartitionSizing)
+	if err != nil {
+		return fmt.Errorf("out %v outstring %v err %v", out, outString, err)
+	}
+
+	currentSize := getPartitionSizing["Size"]
+
+	//if the partition's size is already the size we want this is a noop, just return
+	if currentSize >= finalSize {
+		return nil
```
andyzhangx, line 131: add a logging here?

internal/os/volume/api.go
```diff
@@ -109,6 +110,27 @@ func (VolAPIImplementor) ResizeVolume(volumeID string, size int64) error {
 		finalSize = size
 	}
 
+	cmd = fmt.Sprintf("Get-Volume -UniqueId \"%s\" | Get-partition | Select Size", volumeID)
```
andyzhangx, line 113: could you add a separate func like `func getVolumeSize(id) (int64, error)`, that will make the logic more clear.

andyzhangx (commented):
/ok-to-test

internal/os/volume/api.go
```diff
@@ -109,6 +110,27 @@ func (VolAPIImplementor) ResizeVolume(volumeID string, size int64) error {
...
+	out, err = runExec(cmd)
+
+	if err != nil || len(out) == 0 {
+		return fmt.Errorf("error getting size of the partition from moun. cmd %s, output: %s, error: %v", cmd, string(out), err)
+	}
+
+	var getPartitionSizing map[string]int64
+	outString = string(out)
+	err = json.Unmarshal([]byte(outString), &getPartitionSizing)
```
jingxu97, line 122: instead of using Unmarshal, could we use e.g., 
```
 (Get-Volume -UniqueId \"%s\"  | Get-Partition).Size
strings.TrimSpace(string(out))
```

internal/os/volume/api.go
manueltellez: so far no logging has been done on any of the api in internal/os, I am unsure if that would break a convention

internal/os/volume/api.go
manueltellez: done

ddebroy:
hey can you please squash the commits @manueltellez

manueltellez:
well i did quite a mess trying to clean up the commit. but i think i finally got it right. sorry about that.

PR #97, opened 2020-10-18
Add service commands to system api group

jmpfar:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Adding methods that can start and stop services as well as query some general service information.
This is needed for the iSCSI implementation as iSCSI commands depend on the MSiSCSI service running, and it is stopped by default.

Fixes #

Adding the methods to the still unreleased system/v1alpha1 . If this is an issue this can be moved to a different version.

```release-note
Adding commands that can start/stop/query services to the system group
```


internal/os/system/api.go
```diff
@@ -37,3 +50,51 @@ func (APIImplementor) GetBIOSSerialNumber() (string, error) {
...
+	err = json.Unmarshal(out, &serviceInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	return &serviceInfo, nil
+}
+
+func (APIImplementor) StartService(name string) error {
+	script := `Start-Service -Name $env:ServiceName`
```
ddebroy, line 75: Is there a reason to set the `servicename` parameter through an environment variable here? Why not use `sprintf` directly here rather than a couple of lines below?

internal/os/system/api.go
```diff
@@ -37,3 +50,51 @@ func (APIImplementor) GetBIOSSerialNumber() (string, error) {
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error starting service name=%s. cmd: %s, output: %s, error: %v", name, cmd, string(out), err)
+	}
+
+	return nil
+}
+
+func (APIImplementor) StopService(name string, force bool) error {
+	script := `Stop-Service -Name $env:ServiceName -Force:$([System.Convert]::ToBoolean($env:Force))`
```
ddebroy, line 88: Same comment as before: Is there a reason to set the `servicename` and `force` parameters through environment variables here? Why not use sprintf directly rather than a couple of lines below?

ddebroy (commented):
Looks great! Couple of comments on why environment variables are used to configure the parameters to the cmdlets rather than `sprintf` directly when setting up the command line. If it can be simplified to not use env vars, please address it in a follow-up PR.

ddebroy:
@jmpfar can you also fix update the vendoring and proto newline issue please as noted in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-proxy/97/pull-kubernetes-csi-csi-proxy/1318079150182895616?

internal/os/system/api.go
jmpfar: Hey, thanks for the review! :)

I've built this upon the comments that were left in the smb api group:

https://github.com/kubernetes-csi/csi-proxy/blob/69666768532484464908e227950c9fff6f199389/internal/os/smb/api.go#L55-L60

I tried to find more data online about Powershell command injection, but I haven't found anyone referring to env variables. I did find this [link](https://devblogs.microsoft.com/powershell/protecting-against-malicious-c) which might indicate that env variables could prevent injection. 
Anyway, if this really solves injection then it's a good idea to add this in all csi-proxy commands

internal/os/system/api.go
jmpfar: see above, tell me what you think

internal/os/system/api.go
ddebroy: Thanks for pointing out the link! Makes complete sense to use this protection in case the csi-proxy is being used to invoke arbitrary commands in a malicious context on the host.

PR #99, opened 2020-10-20
Adding the iSCSI APIs

jmpfar:
**What type of PR is this?**
/kind feature

Adding the iSCSI API group to csi-proxy. 
This includes:
* CHAP support
* Limited Mutual CHAP support (secret needs to be set by the user on the node, not through CSIProxy)
* Limited Multipath support (needs to be configured and installed by the user on the node not through CSI Proxy)

Fixes #

* The tests are enabled through an environment variable `ENABLE_ISCSI_TESTS` which needs to be set to `"TRUE"` in order to run the iSCSI tests. This was done as a fail safe mechanism as the integration tests install an iSCSI Target on the local machine and format an empty iSCSI drive. So it's advisable to run this on a clean VM if possible.
* The tests require a Windows Server machine in order to use the iSCSI target feature.

```release-note
Added the iSCSI api group with CHAP support
```


jmpfar:
I've currently disabled the iSCSI integration tests on github workflow as they do not work on that environment.
A possible long term solution is using an image that already has an iSCSI target installed.

Notes to reviewer: 
* The integration tests can be tested on a clean VM without any required prerequisites except for the env var above
* Windows iSCSI target has some issues with long computer names, if you see issues try a VM with a short computer name (<= 15 chars)

jmpfar:
Hey,
I'm trying to rerun the integration tests as they seemed to flake on the last run, any idea how do I trigger those on my pull request?

ddebroy:
@jmpfar invoked `rerun` from the github actions UI linked from the `details` link.

internal/os/iscsi/api.go
```diff
@@ -0,0 +1,192 @@
...
+
+type APIImplementor struct{}
+
+func New() APIImplementor {
+	return APIImplementor{}
+}
+
+func (APIImplementor) AddTargetPortal(portal *TargetPortal) error {
+	cmdLine := fmt.Sprintf(
+		`New-IscsiTargetPortal -TargetPortalAddress ${Env:iscsi_tp_address} ` +
```
ddebroy, line 23: The lack of `-InitiatorInstanceName` parameter assumes csi-proxy will always use the software iSCSI initiator in Windows rather any hardware NICs with iSCSI offload/support - which is fine. If you can add a comment to specify this, that would be great!

ddebroy (commented):
Looks pretty good from a first scan. A few minor comments to clarify code comments.

internal/os/iscsi/api.go
```diff
@@ -0,0 +1,192 @@
...
+func (APIImplementor) ConnectTarget(portal *TargetPortal, iqn string, isMultipath bool,
+	authType string, chapUser string, chapSecret string) error {
+	// Not using InputObject as Connect-IscsiTarget's InputObject does not work.
+	// This is due to being a static WMI method together with a bug in the
+	// powershell version of the API.
+	cmdLine := fmt.Sprintf(
+		`Connect-IscsiTarget -TargetPortalAddress ${Env:iscsi_tp_address}` +
+			` -TargetPortalPortNumber ${Env:iscsi_tp_port} -NodeAddress ${Env:iscsi_target_iqn}` +
+			` -AuthenticationType ${Env:iscsi_auth_type}` +
+			` -IsMultipathEnabled $([System.Convert]::ToBoolean(${Env:iscsi_is_multipath}))`)
```
ddebroy, line 112: If `IsMultipathEnabled` is `true` then `Connect-IscsiTarget` will need to be invoked with `InitiatorPortalAddress` set to the IP address of each network adapter that wants to participate in the multi-path configuration. I think it will probably be better to block specification of `IsMultipathEnabled` for now?

client/api/iscsi/v1alpha1/api.proto
```diff
@@ -0,0 +1,149 @@
...
+  repeated TargetPortal target_portals = 1;
+}
+
+enum AuthenticationType {
+  NONE = 0;
+  ONE_WAY_CHAP = 1;
+  MUTUAL_CHAP = 2;
+}
+
+message ConnectTargetRequest {
```
ddebroy, line 93: Any thoughts on adding a `initiator_address` parameter to `ConnectTargetRequest` ? Can be added in a later version but was wondering if it would be useful in nodes with multiple NICs. We will probably need to another API group potentially as part of the `system` group to query the set of host IP addresses as well.

client/api/iscsi/v1alpha1/api.proto
```diff
@@ -0,0 +1,149 @@
...
+  // needs to be installed as well as MPIO should be correctly configured
+  bool is_multipath = 6;
+}
+
+message ConnectTargetResponse {
+  // Intentionally empty
+}
+
+message GetTargetDisksRequest {
+  // Target portal to which the initiator will connect.
```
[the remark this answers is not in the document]
jmpfar: Fixed

client/api/iscsi/v1alpha1/api.proto
```diff
@@ -0,0 +1,149 @@
...
+}
+
+message GetTargetDisksResponse {
+  // List composed of disk ids (numbers) that are associated with the
+  // iSCSI target
+  repeated string diskIds = 1;
+}
+
+message DisconnectTargetRequest {
+  // Target portal to which the initiator will connect.
```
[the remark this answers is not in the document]
jmpfar: Fixed

client/api/iscsi/v1alpha1/api.proto
```diff
@@ -0,0 +1,149 @@
...
+  TargetPortal target_portal = 1;
+
+  // IQN of the iSCSI Target
+  string iqn = 2;
+}
+
+message GetTargetDisksResponse {
+  // List composed of disk ids (numbers) that are associated with the
+  // iSCSI target
+  repeated string diskIds = 1;
```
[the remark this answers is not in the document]
jmpfar: Fixed

internal/os/iscsi/api.go
jmpfar: I've added clarifying comments (in the proto), I think this is a good idea to add this in later versions.
According to Windows documentation, if you don't specify an initiator:
> "iSCSI initiator service will use any initiator that can reach the target portal."

([AddIScsiSendTargetPortalW](https://docs.microsoft.com/en-us/windows/win32/api/iscsidsc/nf-iscsidsc-addiscsisendtargetportalw))

So, at least by wording, this might suggest that the hardware initiator may be used

internal/os/iscsi/api.go
jmpfar: I didn't consider the multipath scenario where you use two initiators to the same target, what I was thinking about was the other direction
e.g. I have one NIC that connects to two IPs of the storage array,  each one exposing the same IQN.
This scenario will work with the current API
I guess this is more of a scenario for high availability rather than performance, and it was at least what we used in a product I've worked on (although it was far from the normal storage product)

WDYT? The multipath option here was added really quickly just to give an option for users with an already well configured node to use multipath. Full support will probably require more effort anyway, so I can remove the multipath bool and save it for a later version

client/api/iscsi/v1alpha1/api.proto
jmpfar: See the MPIO comments, but I agree, this is a good feature for a later version
This is a mammoth commit as it is, I think I'll start adding features in separate PRs, where the first one is mutual chap support (the node-level set chap function)

internal/os/iscsi/api.go
ddebroy: Interesting - most production multipath configurations I have seen involve redundant paths all the way from the initiator NICs to the target.

I would say let’s remove the multipath option for now in v1alpha1 and add it later if we hear requests/feedback around it.

ddebroy: Once the multipath option is removed for now, I will lgtm/approve the PR. Thanks for getting basic iSCSI support going!

jmpfar: Thanks! 
Removed the multipath option

ddebroy:
Thanks! One more request: can you please squash the commits please? Would like the commit history to not have “address cr comments” commits.

ddebroy:
I squashed and merged this manually. So please ignore https://github.com/kubernetes-csi/csi-proxy/pull/99#issuecomment-715613315 for this PR.

PR #102, opened 2020-10-25
Adding SetMutualChapSecret method to iSCSI API

jmpfar:
**What type of PR is this?**
/kind feature

Adding a method to set the mutual chap secret (see [here](https://docs.microsoft.com/en-us/windows/win32/api/iscsidsc/nf-iscsidsc-setiscsiinitiatorchapsharedsecret)). This is a node method and should be used with care to avoid concurrency issues with other plugins.
I have opted to place it in the iscsi apigroup and not in the system api group. We can use the system group as a place for all methods that affect global state, but I think a better use for system is a place for all os-related functions (version, machine name, etc), rather than methods affecting global state.
I think the iscsi api group is a better place as it is a method that relates to iscsi, but I think this can really go either way

Fixes #

As with other iSCSI api group testing, you will need to set `$env:ENABLE_ISCSI_TESTS = "TRUE"` and run the test on a clean disposable Windows Server VM

```release-note
Adding SetMutualChapSecret to iSCSI API
```


ddebroy (commented):
Adding `SetMutualChapSecret` in the iSCSI API group sounds fine as it's very iSCSI specific. Since we just released the beta.2 version of the client and binary, can you please bump the iSCSI API group version to v1alpha2 and introduce new iSCSI APIs as part of that version?

jmpfar:
Sure, updated commit with v1alpha2 api

PR #123, opened 2021-03-29
Fix issue in isVolumeFormatted function

jingxu97:
Remove the check for "ntfs" during isVolumeFormatted to void mistakenly
format a disk that is formatted with other types of file systems

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/internal/os/volume/api.go
+++ b/internal/os/volume/api.go
@@ -76,7 +76,7 @@ func (VolAPIImplementor) IsVolumeFormatted(volumeID string) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("error checking if volume is formatted. cmd: %s, output: %s, error: %v", cmd, string(out), err)
 	}
-	if len(out) == 0 || !strings.EqualFold(strings.TrimSpace(string(out)), formatFilesystem) {
+	if len(strings.TrimSpace(string(out))) == 0 {
 		return false, nil
 	}
 	return true, nil


wongma7:
With this change, after I call New-Partition on a new disk, the returned FileSystemTypereturned is "Unknown", not "", so IsVolumeFormatted returns true and my code skips formatting the volume. Before it would return false and the code would format the volume as expected.

```
PS C:\Users\Administrator> (Get-Volume -UniqueId "\\?\Volume{cb32b357-0000-0000-0000-010000000000}\" -ErrorAction Stop).FileSystemType
Unknown
```
In GUI the disk shows as having "RAW" partition.

How should the "Unknown" case be handled?

jingxu97:
> With this change, after I call New-Partition on a new disk, the returned FileSystemType* is "Unknown", not "", so IsVolumeFormatted returns true and my code skips formatting the volume. Before it would return false and the code would format the volume as expected.
> 
> ```
> PS C:\Users\Administrator> (Get-Volume -UniqueId "\\?\Volume{cb32b357-0000-0000-0000-010000000000}\" -ErrorAction Stop).FileSystemType
> Unknown
> ```
> 
> In GUI the disk shows as having "RAW" partition.
> 
> How should the "Unknown" case be handled?

Thanks, @wongma7 I will fix it.

PR #126, opened 2021-04-14
Close Open disk file handles

wongma7:
**What type of PR is this?**
/kind bug

If csi-proxy keeps these files open it blocks EBS detach from succeeding.

I tested this fix and it works for me.

Full story:

While debugging my EBS CSI implementation I found everything worked fine except for volume Detach. EBS support said OS was reporting something holding on to the disk (despite me trying removing all accesspoints, taking disk offline, etc.). Eventually after much guessing and testing, narrowed it to csi-proxy, i.e. as soon as I stopped it Detach would proceed. (I'm no Windows expert so possibly there was an easier way to find out what was holding onto device, like via process explorer, but yeah...)

Fixes #

```release-note
Fix GetDiskNumberByName and ListDiskIDs keeping file handles open on disks
```

### DIFF
--- a/internal/os/disk/api.go
+++ b/internal/os/disk/api.go
@@ -5,6 +5,7 @@ package disk
 import (
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os/exec"
 	"regexp"
@@ -253,18 +254,34 @@ func (imp APIImplementor) GetDiskNumberWithID(page83ID string) (uint32, error) {
 	json.Unmarshal([]byte(outString), &disks)
 
 	for i := range disks {
-		h, err := syscall.Open(disks[i].Path, syscall.O_RDONLY, 0)
-		if err != nil {
+		diskNumber, err := imp.getDiskNumberByID(disks[i].Path, page83ID)
+		if err == ErrNotFound {
+			continue
+		} else if err != nil {
 			return 0, err
-		}
-
-		found, err := imp.DiskHasPage83ID(h, page83ID)
-		if found {
-			return imp.GetDiskNumber(h)
+		} else {
+			return diskNumber, nil
 		}
 	}
 
 	return 0, fmt.Errorf("Could not find disk with Page83 ID %s", page83ID)
+}
+
+var ErrNotFound = errors.New("not found")
+
+func (imp APIImplementor) getDiskNumberByID(path string, page83ID string) (uint32, error) {
+	h, err := syscall.Open(path, syscall.O_RDONLY, 0)
+	defer syscall.Close(h)
+	if err != nil {
+		return 0, err
+	}
+
+	found, err := imp.DiskHasPage83ID(h, page83ID)
+	if found {
+		return imp.GetDiskNumber(h)
+	}
+
+	return 0, ErrNotFound
 }
 
 // ListDiskIDs - constructs a map with the disk number as the key and the DiskID structure
@@ -282,19 +299,9 @@ func (imp APIImplementor) ListDiskIDs() (map[string]shared.DiskIDs, error) {
 	m := make(map[string]shared.DiskIDs)
 
 	for i := range disks {
-		h, err := syscall.Open(disks[i].Path, syscall.O_RDONLY, 0)
+		diskNumber, page83, err := imp.getDiskNumberAndIDByPath(disks[i].Path)
 		if err != nil {
 			return nil, err
-		}
-
-		page83, err := imp.GetDiskPage83ID(h)
-		if err != nil {
-			return m, fmt.Errorf("Could not get page83 ID: %v", err)
-		}
-
-		diskNumber, err := imp.GetDiskNumber(h)
-		if err != nil {
-			return m, fmt.Errorf("Could not get disk number: %v", err)
 		}
 
 		diskNumString := strconv.FormatUint(uint64(diskNumber), 10)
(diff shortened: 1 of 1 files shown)


internal/os/disk/api.go
```diff
@@ -262,6 +262,11 @@ func (imp APIImplementor) GetDiskNumberWithID(page83ID string) (uint32, error) {
 		if found {
 			return imp.GetDiskNumber(h)
 		}
+
+		err = syscall.Close(h)
```
andyzhangx, line 266: what about use `defer` after `syscall.Open`?

jingxu97:
@wongma7 could you address the comment?

internal/os/disk/api.go
wongma7: sure, updated!

wongma7:
/assign @msau42

please approve!  2 liner PR! many thanks!

internal/os/disk/api.go
```diff
@@ -254,6 +254,7 @@ func (imp APIImplementor) GetDiskNumberWithID(page83ID string) (uint32, error) {
 
 	for i := range disks {
 		h, err := syscall.Open(disks[i].Path, syscall.O_RDONLY, 0)
+		defer syscall.Close(h)
```
msau42, line 257: What's the maximum number of disks a node can have? Would it be worth enclosing this loop in a code section so that it gets closed at the end of each loop?

wongma7: for ec2 depends on instance type but typically like 20-30
https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/volume_limits.html#instance-type-volume-limits (EKS only supports windows nitro instances so the "AWS PV" driver in the other section don't apply)

do you mean to wrap it like this? I don't think it's worth the loss in readability 
```
func() (uint32, error) {
   ...
   syscall.Open
   defer syscall.Close(h)
}()
```

msau42: Or a separate function could also work.

The default fd limit per process is 1024 for Linux. I don't know about Windows. Does Windows also have a maximum disk limit? cc @ddebroy @jingxu97

mauriciopoppe: here it says that it's 16M https://docs.microsoft.com/en-us/archive/blogs/markrussinovich/pushing-the-limits-of-windows-handles#maximum-number-of-handles

wongma7: did some refactoring so we always open and close 1 handle at a time

wongma7:
/hold

need to retest on my setup

jingxu97:
@wongma7 thanks for this PR. Pls let me know after you test it.

wongma7:
/hold cancel

together with https://github.com/kubernetes-csi/csi-proxy/pull/145 I tested that attach, mount, unmount, detach works with ebs

internal/os/disk/api.go
```diff
@@ -308,6 +315,26 @@ func (imp APIImplementor) ListDiskIDs() (map[string]shared.DiskIDs, error) {
 	return m, nil
 }
 
+func (imp APIImplementor) getPathDiskNumberAndID(path string) (uint32, string, error) {
```
jingxu97, line 318: getDiskNumerAndIDByPath?

internal/os/disk/api.go
```diff
@@ -253,20 +254,36 @@ func (imp APIImplementor) GetDiskNumberWithID(page83ID string) (uint32, error) {
...
+			return diskNumber, nil
 		}
 	}
 
 	return 0, fmt.Errorf("Could not find disk with Page83 ID %s", page83ID)
 }
 
+var ErrNotFound = errors.New("not found")
+
+func (imp APIImplementor) getPathDiskNumberWithID(path string, page83ID string) (uint32, error) {
```
jingxu97, line 272: GetDiskNumberByID?

internal/os/disk/api.go
wongma7: fixed, much better...

internal/os/disk/api.go
wongma7: fixed

internal/os/disk/api.go
```diff
@@ -253,20 +254,36 @@ func (imp APIImplementor) GetDiskNumberWithID(page83ID string) (uint32, error) {
...
+			return diskNumber, nil
 		}
 	}
 
 	return 0, fmt.Errorf("Could not find disk with Page83 ID %s", page83ID)
 }
 
+var ErrNotFound = errors.New("not found")
+
+func (imp APIImplementor) getDiskNumberByID(path string, page83ID string) (uint32, error) {
```
jingxu97, line 272: I am kind of confused by these two functions getDiskNumberAndIDByPath and getDiskNumberByID. 

Is there a way to kind of combine those?

jingxu97:
@mauriciopoppe is working on disk API part https://github.com/kubernetes-csi/csi-proxy/pull/140
wondering whether you want to wait for that PR merge since there might be some name changes?

internal/os/disk/api.go
wongma7: okay, let me combine. 

(the PR started as 2 lines, bit by bit it has ballooned, so I am getting confused too :joy:)

wongma7:
@jingxu97 I'm going to revert to the original PR where I just call syscall.Close(h) at end of the loop. I don't want to refactor and then rebase if beta3 API will make it obsolete

PR #127, opened 2021-04-14
Fix IsVolumeFormatted issue on checking volume type "Unknown"

jingxu97:
This PR fixes issue in IsVolumeFormatted on checking volume type
"Unknown"

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/internal/os/volume/api.go
+++ b/internal/os/volume/api.go
@@ -76,7 +76,8 @@ func (VolAPIImplementor) IsVolumeFormatted(volumeID string) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("error checking if volume is formatted. cmd: %s, output: %s, error: %v", cmd, string(out), err)
 	}
-	if len(strings.TrimSpace(string(out))) == 0 {
+	stringOut := strings.TrimSpace(string(out))
+	if len(stringOut) == 0 || strings.EqualFold(stringOut, "Unknown") {
 		return false, nil
 	}
 	return true, nil


wongma7:
thx so much for quick fix, I just tested this patch and can confirm it works for me. I added a little log statement to verify that my "raw" volumes are showing up with `Unknown` FS for me and matching the new condition:

W0415 19:13:08.177499   10684 api.go:82] STRING "Unknown"

```
diff --git a/internal/os/volume/api.go b/internal/os/volume/api.go
index 74f0401..afdcf3b 100644
--- a/internal/os/volume/api.go
+++ b/internal/os/volume/api.go
@@ -7,6 +7,8 @@ import (
        "regexp"
        "strconv"
        "strings"
+
+       "k8s.io/klog/v2"
 )
 
 const formatFilesystem = "ntfs"
@@ -76,7 +78,9 @@ func (VolAPIImplementor) IsVolumeFormatted(volumeID string) (bool, error) {
        if err != nil {
                return false, fmt.Errorf("error checking if volume is formatted. cmd: %s, output: %s, error: %v", cmd, string(out), err)
        }
-       if len(strings.TrimSpace(string(out))) == 0 {
+       stringOut := strings.TrimSpace(string(out))
+       klog.Warningf("STRING %q", stringOut)
+       if len(stringOut) == 0 || strings.EqualFold(stringOut, "Unknown") {
                return false, nil
        }
        return true, nil
```

andyzhangx (approved):
/lgtm

PR #147, opened 2021-06-03
Close Open disk file handles

wongma7:
**What type of PR is this?**
/kind bug

this is  a continuation/refactor of https://github.com/kubernetes-csi/csi-proxy/pull/126

@jingxu97 PTAL. as you suggested I combined the two confusing functions into one, rebased on the v1beta3 PR, then retested everything and it works! i.e. EBS detach succeeds!

Fixes #

```release-note
NONE
```

### DIFF
--- a/internal/os/disk/api.go
+++ b/internal/os/disk/api.go
@@ -194,52 +194,6 @@ func (DiskAPI) GetDiskNumber(disk syscall.Handle) (uint32, error) {
 	return devNum.DeviceNumber, err
 }
 
-func (DiskAPI) DiskHasPage83ID(disk syscall.Handle, matchID string) (bool, error) {
-	query := StoragePropertyQuery{}
-
-	bufferSize := uint32(4 * 1024)
-	buffer := make([]byte, 4*1024)
-	var size uint32
-	var n uint32
-	var m uint16
-
-	query.QueryType = PropertyStandardQuery
-	query.PropertyID = StorageDeviceIDProperty
-
-	querySize := uint32(unsafe.Sizeof(query.PropertyID)) + uint32(unsafe.Sizeof(query.QueryType)) + uint32(unsafe.Sizeof(query.Byte))
-	querySize = uint32(unsafe.Sizeof(query))
-	err := syscall.DeviceIoControl(disk, IOCTL_STORAGE_QUERY_PROPERTY, (*byte)(unsafe.Pointer(&query)), querySize, (*byte)(unsafe.Pointer(&buffer[0])), bufferSize, &size, nil)
-	if err != nil {
-		return false, fmt.Errorf("IOCTL_STORAGE_QUERY_PROPERTY failed: %v", err)
-	}
-
-	devIDDesc := (*StorageDeviceIDDescriptor)(unsafe.Pointer(&buffer[0]))
-
-	pID := (*StorageIdentifier)(unsafe.Pointer(&devIDDesc.Identifiers[0]))
-
-	page83ID := []byte{}
-	byteSize := unsafe.Sizeof(byte(0))
-	for n = 0; n < devIDDesc.NumberOfIdentifiers; n++ {
-		if pID.Association == StorageIDAssocDevice && (pID.CodeSet == StorageIDCodeSetBinary || pID.CodeSet == StorageIDCodeSetASCII) {
-			for m = 0; m < pID.IdentifierSize; m++ {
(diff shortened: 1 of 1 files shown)


internal/os/disk/api.go
```diff
@@ -194,52 +194,6 @@ func (DiskAPI) GetDiskNumber(disk syscall.Handle) (uint32, error) {
...
-	}
-
-	devIDDesc := (*StorageDeviceIDDescriptor)(unsafe.Pointer(&buffer[0]))
-
-	pID := (*StorageIdentifier)(unsafe.Pointer(&devIDDesc.Identifiers[0]))
-
-	page83ID := []byte{}
-	byteSize := unsafe.Sizeof(byte(0))
-	for n = 0; n < devIDDesc.NumberOfIdentifiers; n++ {
-		if pID.Association == StorageIDAssocDevice && (pID.CodeSet == StorageIDCodeSetBinary || pID.CodeSet == StorageIDCodeSetASCII) {
```
mauriciopoppe, line 223: I saw that this line is different in `GetDiskPage83ID`, @ddebroy might help us here

internal/os/disk/api.go
```diff
@@ -194,52 +194,6 @@ func (DiskAPI) GetDiskNumber(disk syscall.Handle) (uint32, error) {
...
-			for m = 0; m < pID.IdentifierSize; m++ {
-				page83ID = append(page83ID, *(*byte)(unsafe.Pointer(uintptr(unsafe.Pointer(&pID.Identifier[0])) + byteSize*uintptr(m))))
```
mauriciopoppe, line 225: `page83ID` in this will only be updated if `pID.CodeSet == StorageIDCodeSetBinary || pID.CodeSet == StorageIDCodeSetASCII`, however in `GetDiskPage83ID` it's updated without this condition, I think that's the only difference between these two methods

wongma7: good catch, it looked like just an optimization to me to check CodeSet early because, if the CodeSet is neither, there is no way that the function will find the page83ID there. behavior should be the same with or without the condition

wongma7: never mind, it looks like codeset can also be "UTF8" or "reserved"? probably better to have the condition.

internal/os/disk/api.go
```diff
@@ -294,20 +248,39 @@ func (imp DiskAPI) GetDiskNumberWithID(page83ID string) (uint32, error) {
...
+func (imp DiskAPI) GetDiskNumberAndPage83ID(path string) (uint32, string, error) {
+	h, err := syscall.Open(path, syscall.O_RDONLY, 0)
+	defer syscall.Close(h)
+	if err != nil {
+		return 0, "", err
+	}
+
+	diskNumber, err := imp.GetDiskNumber(h)
+	if err != nil {
+		return 0, "", err
```
ddebroy, line 273: A DiskNumber of `0` is a valid ID for a disk enumerated as \\.\PhysicalDrive0. So returning 0 for the error scenario is not recommended. Maybe return a pointer to uint32 or -1?

ddebroy (commented):
LGTM with one suggestion

internal/os/disk/api.go
wongma7: hmm I would expect caller to always check the error first. -1 i can't do since it's unsigned (or caller needs to convert to unsigned), pointer maybe I could do but then the caller needs to always check if it's nil before dereference, and I would argue in that case they should just check the error

ddebroy: Looks like in other functions we return 0 in the error case too. So makes sense to check error first.

internal/os/disk/api.go
ddebroy: I think we restricted with the extra conditions for Ascii and Binary because those are the ones that were handled in the conversions with `string` and `hex.EncodeToString`. Extending it to `utf8` would be nice. For `Reserved`, we have no idea how to handle the formatting.

wongma7: I added the conditions back just to be safe. https://github.com/kubernetes-csi/csi-proxy/pull/147/files#diff-8dec85397edab1ec05fe2a8d6f36c534b7cd19dd8d6db9010b75536f597a54a6R223

I guess Reserved means it is reserved by spec and should never ever be seen in practice. i suppose utf8 might in the future but, at least for ebs disks none have utf8, i'll leave it out of scope of this PR

PR #152, opened 2021-06-15
Rename /internal to pkg, nested /internal/server/<group>/internal to impl

mauriciopoppe:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
/kind cleanup

**What this PR does / why we need it**:

For the upcoming move to Windows privileged containers we need to make some changes in the file structure of the codebase, `internal` folders can't be imported from other projects so we need to rename them, the new names are:

- `/internal` to `/pkg`
- `/internal/server/<group>/internal` to `/pkg/server/<group>/impl`

I also changed the code generator to create the generated files at the folders above

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

/cc @jingxu97 @ddebroy

PR #154, opened 2021-06-15
Send arrays to ConvertTo-Json in the internal disk implementation

### LINKED ISSUE #153: CSI-Proxy fails to return the disks through disk.ListDiskIDs and disk.ListDiskLocations in a VM with only one disk
**What happened**:

In a VM with only a single disk (like when it's a new VM) the integration tests fail while trying to get the list of disks, this command:

```
(Get-Disk | Select Path, SerialNumber) | ConvertTo-Json
```

Will return an object if there's only one disk but an array of objects if there are many disks, the go code is always unmarshalling an array so if it only has one disk it fails during unmarshalling (silently), the result is that this method returns no disks when there's one disk.

https://github.com/kubernetes-csi/csi-proxy/blob/e7c1df0223edc2d12e6b1e42d6251686018ef57e/internal/os/disk/api.go#L286-L305

mauriciopoppe:
**What type of PR is this?**
>
/kind bug

Context #153, we're using `ConvertTo-Json @(...)` instead of pipes

Fixes #153 

```release-note
NONE
```

/hold
/cc @jingxu97 @ddebroy 

We should merge #152 first

### DIFF
--- a/pkg/os/disk/api.go
+++ b/pkg/os/disk/api.go
@@ -79,7 +79,7 @@ func (DiskAPI) ListDiskLocations() (map[uint32]shared.DiskLocation, error) {
 	//    "number":  0,
 	//    "location":  "PCI Slot 3 : Adapter 0 : Port 0 : Target 1 : LUN 0"
 	// }, ...]
-	cmd := fmt.Sprintf("Get-Disk | select number, location | ConvertTo-Json")
+	cmd := fmt.Sprintf("ConvertTo-Json @(Get-Disk | select Number, Location)")
 	out, err := runExec(cmd)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list disk location. cmd: %q, output: %q, err %v", cmd, string(out), err)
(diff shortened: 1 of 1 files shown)


mauriciopoppe:
I've also checked the other places in `volume` that were using `| ConvertTo-Json` and they don't use array, they use maps so I think we're ok

/unhold

pkg/os/disk/api.go
```diff
@@ -93,8 +93,8 @@ func (DiskAPI) ListDiskLocations() (map[uint32]shared.DiskLocation, error) {
 
 	m := make(map[uint32]shared.DiskLocation)
 	for _, v := range getDisk {
-		str := v["location"].(string)
-		num := v["number"].(float64)
+		str := v["Location"].(string)
```
jingxu97, line 96: why change location --> Location?

mauriciopoppe: this change is actually not needed and I can revert it back, it's because I also changed this line above:

```
cmd := fmt.Sprintf("ConvertTo-Json @(Get-Disk | select Number, Location)")
```

mauriciopoppe: I think I can skip making this change, I already run the integration tests on #155 and they're passing with the v1 API too

jingxu97:
/lgtm
approve

PR #174, opened 2021-09-16
Added file descriptor to hide password

### LINKED ISSUE #142: Password field should be hidden in the SMB NewSmbGlobalMappingRequest Proto
This field should be hidden https://github.com/kubernetes-csi/csi-proxy/blob/c2495301dc4209759d986af0c40b5e1584f71866/client/api/smb/v1beta2/api.proto#L34

- [ ] Investigate how to update csi-proxy's build process to include additional proto plugins like in https://github.com/container-storage-interface/spec/blob/da58351ba3d7baf850877be3425e76ff30645d33/lib/go/Makefile#L134-L144
- [ ] Extend the FieldOptions field like https://github.com/container-storage-interface/spec/blob/da58351ba3d7baf850877be3425e76ff30645d33/csi.proto#L23-L32
- [ ] Use the field option in the `password` field

Kartik494:
Motivation of this PR is to hide password from logging and debugging

**What type of PR is this?**
 /kind bug

Fixes #142


Kartik494:
cc @mauriciopoppe  Is there any tooling in our k8s repo to generate pb.go file for corresponding api changes?

mauriciopoppe:
@Kartik494 yes, please read docs/DEVELOPMENT.md to recreate the pb.go files

Kartik494:
@mauriciopoppe Thanks for suggestion, i will go through docs/DEVELOPMENT.md and update the pb.go file

Kartik494:
Hi @mauriciopoppe I have tried generating corrseponding pb.go files as [documented](https://github.com/kubernetes-csi/csi-proxy/blob/master/docs/DEVELOPMENT.md#making-changes-to-the-client-module) . As documented warning and errors are expected but i am unable to generate  corrseponding pb.go file.

Kartik494:
> Hi @mauriciopoppe I have tried generating corrseponding pb.go files as [documented](https://github.com/kubernetes-csi/csi-proxy/blob/master/docs/DEVELOPMENT.md#making-changes-to-the-client-module) . As documented warning and errors are expected but i am unable to generate corrseponding pb.go file.

It is fixed now , my local env is messed with protoc

Kartik494:
/retitle  Added file descriptor to hide password

client/api/smb/v1beta2/api.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.26.0
```
mauriciopoppe, line 3: There's this rule in the repo that pb.go files that were already released shouldn't be modified, because smb already reached v1 I guess that we'd need to create a new v2alpha1 API version, cc @ddebroy

client/api/smb/v1beta2/api.proto
```diff
@@ -2,8 +2,21 @@ syntax = "proto3";
...
 
+extend google.protobuf.FieldOptions {
+    // Indicates that a field MAY contain information that is sensitive
+    // and MUST be treated as such (e.g. not logged).
+    bool csi_secret = 1059;
+
+    // Indicates that this field is OPTIONAL and part of an experimental
+    // API that may be deprecated and eventually removed between minor
+    // releases.
+    bool alpha_field = 1060;
```
mauriciopoppe, line 17: it looks like we aren't using alpha_field, only csi_secret

mauriciopoppe (commented):
Thanks for working on this, it looks like `csi_secret` is just an annotation on a property of a message, how is CSI Spec using it? is it leaving it to the users of the CSI Spec library to decide whether to log a secret field or not?

client/api/smb/v1beta2/api.pb.go
Kartik494: Actually i am not able to generate respective pb.go file so when i used this version i am able to do so. That's why i have used this version? do i need to use 1.26v only?

client/api/smb/v1beta2/api.proto
Kartik494: In this `fieldOption` `bool aplha_field` is optional field  i don't have exact idea on this if this not impact changes, so should i remove this?

client/api/smb/v1beta2/api.pb.go
mauriciopoppe: I think that using the latest version of protoc-gen-go works however my point is about the file that's getting modified, @ddebroy said (and it's also pointed out in [the docs](https://github.com/kubernetes-csi/csi-proxy/blob/master/docs/DEVELOPMENT.md#making-changes-to-the-client-module)) that once the api.pb.go file of a version is released then it shouldn't be modified, we already released smb/v1beta2 so I think it should stay as it is.

Instead we could:
- bump the SMB version from v1 to v2alpha1
- make the changes in the proto file like adding the csi secrets annotation
- generate `client/api/smb/v2alpha/api.pb.go`

I'm still not sure how we'd use the csi secrets annotation, to check it I would: 

- see where it's used in the CSI Spec (which methods and messages)
- read the CSI sidecar or the CSI Driver code that's using the method/message and see what it's doing with the field, is it checking that the field has a property to do special logging?

client/api/smb/v1beta2/api.proto
mauriciopoppe: I think that it's referring to the field that's going to be annotated, in this example from the CSI Spec it'd be the `volume_mount_group` field

```
  message MountVolume {
    // The filesystem type. This field is OPTIONAL.
    // An empty string is equal to an unspecified field value.
    string fs_type = 1;

    // The mount options that can be used for the volume. This field is
    // OPTIONAL. `mount_flags` MAY contain sensitive information.
    // Therefore, the CO and the Plugin MUST NOT leak this information
    // to untrusted entities. The total size of this repeated field
    // SHALL NOT exceed 4 KiB.
    repeated string mount_flags = 2;

    // If SP has VOLUME_MOUNT_GROUP node capability and CO provides
    // this field then SP MUST ensure that the volume_mount_group
    // parameter is passed as the group identifier to the underlying
    // operating system mount system call, with the understanding
    // that the set of available mount call parameters and/or
    // mount implementations may vary across operating systems.
    // Additionally, new file and/or directory entries written to
    // the underlying filesystem SHOULD be permission-labeled in such a
    // manner, unless otherwise modified by a workload, that they are
    // both readable and writable by said mount group identifier.
    // This is an OPTIONAL field.
    string volume_mount_group = 3 [(alpha_field) = true];
  }
```

Kartik494: @mauriciopoppe Thanks for your input , i  will addressed this ASAP

mauriciopoppe:
/close

Closing this PR because of the explanation in https://github.com/kubernetes-csi/csi-proxy/issues/142#issuecomment-1034532971, sorry for not finding out this before

Kartik494:
@mauriciopoppe Sure, No problem!

PR #145, opened 2021-06-01
Don't count reserved partitions (gpt) when checking if any exist

wongma7:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:  see https://github.com/kubernetes-csi/csi-proxy/issues/144: with gpt being default, there is some 128MB reserved partition at beginning of disk. LIke
```
PartitionNumber  DriveLetter Offset                                        Size Type
---------------  ----------- ------                                        ---- ----
1                            1048576                                     128 MB Reserved
2                            135266304                                 19.87 GB Basic
```
Don't count it when checking if partition exists.

**Which issue(s) this PR fixes**:

Fixes #144

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #175, opened 2021-09-29
fix: check partition error on MBR disk

andyzhangx:
**What type of PR is this?**
/kind bug

fix: check partition error on MBR disk
Before k8s 1.22, all default windows disk type is MBR, and `Type` of `Get-Partition` for MBR disk is `IFS`, PR(https://github.com/kubernetes-csi/csi-proxy/pull/145) brings the breaking change, it would lead to check partition error on MBR disk, this PR only filters `Basic` type which could solve the issue.

Fixes https://github.com/Azure/AKS/issues/2568

/assign @wongma7 @jingxu97 

```release-note
fix: check partition error on MBR disk
```

### DIFF
--- a/pkg/os/disk/api.go
+++ b/pkg/os/disk/api.go
@@ -158,7 +158,7 @@ func (DiskAPI) InitializeDisk(diskNumber uint32) error {
 }
 
 func (DiskAPI) BasicPartitionsExist(diskNumber uint32) (bool, error) {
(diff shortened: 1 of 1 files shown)


wongma7:
/lgtm

my b, csi-proxy 1.0.0 will always format GPT so excluding reserved is compatible with itself, but that is not backwards compatible with previous MBR volumes

ggriffiths:
Looks like this will be backwards compatible with the previous partition check from before #145.
/lgtm

PR #179, opened 2021-11-05
WIP: build/deploy csi-proxy in host-process container

marosset:
Signed-off-by: Mark Rossetti <marosset@microsoft.com>

**What type of PR is this?**
/kind feature

This PR adds functionality to build a container image containing csi-proxy.exe and deploy it as a daemonSet using [HostProcess containers](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/)

Fixes #

I have testing using csi-proxy deployed in a host-process container in conjunction with azurefile-csi-driver and everything is behaving as expected.

```release-note

```


image/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG  BASE
+
+FROM --platform=linux/amd64 ubuntu:20.04 as linuxbase
```
jsturtevant, line 3: is needed if you are are not doing anything on this stage?

marosset:
I think this should probably get tied into https://github.com/kubernetes-csi/csi-release-tools but I need to formalize myself with that process.

image/build-container-image.sh
```diff
@@ -0,0 +1,33 @@
...
+set -ex
+
+: "${VERSION:?VERISON not set}"
+: "${REGISTRY:?REGISTRY not set}"
+
+export DOCKER_CLI_EXPERIMENTAL=enabled
+if [ $(docker buildx ls | grep -c img-builder) == 0 ]; then
+    docker buildx create --name img-builder 
+fi
+docker buildx use img-builder
```
jsturtevant, line 17: might consider cleaning up the image builder:

`trap 'docker buildx rm img-builder' EXIT`

image/build-container-image.sh
```diff
@@ -0,0 +1,33 @@
...
+
+set -o nounset
+set -ex
+
+: "${VERSION:?VERISON not set}"
+: "${REGISTRY:?REGISTRY not set}"
+
+export DOCKER_CLI_EXPERIMENTAL=enabled
+if [ $(docker buildx ls | grep -c img-builder) == 0 ]; then
+    docker buildx create --name img-builder
```
jsturtevant, line 15: In a recent update to buildx I had to begin explicitly requesting windows:

`docker buildx create --name img-builder --use --platform windows/amd64`

image/build-container-image.sh
```diff
@@ -0,0 +1,33 @@
...
+
+export DOCKER_CLI_EXPERIMENTAL=enabled
+if [ $(docker buildx ls | grep -c img-builder) == 0 ]; then
+    docker buildx create --name img-builder 
+fi
+docker buildx use img-builder
+docker buildx build --platform windows/amd64 --output=type=registry --pull -f Dockerfile --build-arg BASE=mcr.microsoft.com/windows/nanoserver:1809 -t $REGISTRY/csi-proxy:$VERSION-1809 ..
+docker buildx build --platform windows/amd64 --output=type=registry --pull -f Dockerfile --build-arg BASE=mcr.microsoft.com/windows/nanoserver:ltsc2022 -t $REGISTRY/csi-proxy:$VERSION-ltsc2022 ..
+
+docker manifest create $REGISTRY/csi-proxy:$VERSION $REGISTRY/csi-proxy:$VERSION-1809 $REGISTRY/csi-proxy:$VERSION-ltsc2022
```
jsturtevant, line 21: In Kubernetes test images we keep these values as separate config: https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/BASEIMAGE

I guess we only have 2019 and WS 2022 to deal with now so this isn't an issue

image/build-container-image.sh
```diff
@@ -0,0 +1,33 @@
...
+os_version_1809=$(docker manifest inspect mcr.microsoft.com/windows/nanoserver:1809 | grep "os.version" | head -n 1 | awk -F\" '{print $4}')
+docker manifest annotate --os windows --arch amd64 --os-version $os_version_1809 $REGISTRY/csi-proxy:$VERSION $REGISTRY/csi-proxy:$VERSION-1809
+
+os_version_ltsc2022=$(docker manifest inspect mcr.microsoft.com/windows/nanoserver:ltsc2022 | grep "os.version" | head -n 1 | awk -F\" '{print $4}')
+docker manifest annotate --os windows --arch amd64 --os-version $os_version_ltsc2022 $REGISTRY/csi-proxy:$VERSION $REGISTRY/csi-proxy:$VERSION-ltsc2022
+
+docker manifest inspect $REGISTRY/csi-proxy:$VERSION
+docker manifest push $REGISTRY/csi-proxy:$VERSION
+
+
```
jsturtevant, line 32: nit: extra lines

deploy/csi-proxy-ds.yaml
```diff
@@ -0,0 +1,28 @@
...
+      nodeSelector:
+        "kubernetes.io/os": windows
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\SYSTEM"
+      hostNetwork: true
+      containers:
+        - name: csi-proxy
+          image: mrosse3/csi-proxy:v1.0.2
```
mauriciopoppe, line 27: should be `k8s.gcr.io/sig-storage/csi-proxy` before the merge

mauriciopoppe (commented):
This looks interesting, thanks for the PR!

> I think this should probably get tied into kubernetes-csi/csi-release-tools but I need to formalize myself with that process.

yeah, we use `release-tools/build.make`, it already handles creating a multiarch binary too for many Windows platforms, I'd create a Dockerfile.Windows file like https://github.com/kubernetes-csi/node-driver-registrar/blob/master/Dockerfile.Windows and modify the root `Makefile` to call push-multiarch instead

deploy/csi-proxy-ds.yaml
marosset: ack!

image/Dockerfile
marosset: It is not needed. I'll remove it.

deploy/csi-proxy-ds.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: csi-proxy
+  namespace: kube-system # Is this the right namespace?
```
ddebroy, line 5: Looks good based on https://github.com/kubernetes-csi/csi-driver-smb/blob/master/deploy/csi-smb-node.yaml#L6 and https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/deploy/csi-nfs-node.yaml#L8

deploy/csi-proxy-ds.yaml
```diff
@@ -0,0 +1,28 @@
...
+    matchLabels:
+      name: csi-proxy
+  template:
+    metadata:
+      labels:
+        name: csi-proxy
+    spec:
+      tolerations:
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
```
ddebroy, line 17: Windows master nodes are not supported right? Do we need this?

ddebroy (commented):
This is great! It would be nice to align with release-tools and a structure like https://github.com/kubernetes-csi/csi-driver-smb/tree/master/deploy .

marosset:
Thanks all, let me look into aligning with the release-tools.

marosset:
@mauriciopoppe I made some updates and ran `make push-multiarch PULL_BASE_REF=master REGISTRY_NAME=mrosse3 BUILD_PLATFORMS="windows amd64 .exe nanoserver:1809; windows amd64 .exe nanoserver:ltsc2022"`.
Is there anything else I need to do to integrate w/ the release tools?
This was a lot more straight-forward than I was expecting!

mauriciopoppe:
@marosset I've been reading the scripts related with our build process, we have two:

- presubmit: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/csi-proxy/csi-proxy-config.yaml
- postsubmit: https://github.com/kubernetes/test-infra/blob/7180a0f5fab405667c76a7c0a76af9fe260c5c74/config/jobs/image-pushing/k8s-staging-sig-storage.yaml#L559-L595

I think the overall objective is to call the function `gcr_cloud_build` declared in `release-tools/prow.sh`, that function takes care of building a multiarch image, however we shouldn't break the code that generates the standalone binary too as this is another release artifact.

For the presubmit job the file used is `.prow.sh` which is sourcing `release-tools/prow.sh` and calling `run_with_go`, I think that `.prow.sh` should use the function `gcr_cloud_build` instead of the two `run_with_go` calls, `gcr_cloud_build` calls `run_with_go` under the hood.

For the postsubmit job the file used is `cloudbuild.yaml` which is calling `.cloudbuild.sh`, in other CSI repos the file `.cloudbuild.sh` is a symlink to `release-tools/cloudbuild.sh` which is calling `gcr_cloud_build` too, we might make it a symlink or call the same functions as `release-tools/cloudbuild.sh`.

marosset:
> @marosset I've been reading the scripts related with our build process, we have two:
> 
> * presubmit: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/csi-proxy/csi-proxy-config.yaml
> * postsubmit: https://github.com/kubernetes/test-infra/blob/7180a0f5fab405667c76a7c0a76af9fe260c5c74/config/jobs/image-pushing/k8s-staging-sig-storage.yaml#L559-L595
> 
> I think the overall objective is to call the function `gcr_cloud_build` declared in `release-tools/prow.sh`, that function takes care of building a multiarch image, however we shouldn't break the code that generates the standalone binary too as this is another release artifact.
> 
> For the presubmit job the file used is `.prow.sh` which is sourcing `release-tools/prow.sh` and calling `run_with_go`, I think that `.prow.sh` should use the function `gcr_cloud_build` instead of the two `run_with_go` calls, `gcr_cloud_build` calls `run_with_go` under the hood.
> 
> For the postsubmit job the file used is `cloudbuild.yaml` which is calling `.cloudbuild.sh`, in other CSI repos the file `.cloudbuild.sh` is a symlink to `release-tools/cloudbuild.sh` which is calling `gcr_cloud_build` too, we might make it a symlink or call the same functions as `release-tools/cloudbuild.sh`.

Thanks @mauriciopoppe this is very helpful.
Let me make these updates.

.prow.sh
```diff
@@ -14,3 +14,6 @@ ensure_paths
 # main
 run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make all "GOFLAGS_VENDOR=${GOFLAGS_VENDOR}" "BUILD_PLATFORMS=${CSI_PROW_BUILD_PLATFORMS}"
 run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make -k test "GOFLAGS_VENDOR=${GOFLAGS_VENDOR}" 2>&1 | make_test_to_junit
+
+# build / push multi-arch images for validation
+gcr_cloud_build
```
marosset, line 19: gcr_cloud_build just calls make push-multiarch so i left in the calls to run_with_go make all and make test.
Should I update the push-multiarch make target to run the tests too so this file can just call gcr_cloud_build and not the run_with_go commands?
That seems a little convoluted to me.

marosset:
@mauriciopoppe is gcr.io/k8s-staging-sig-storage the correct registry to use for presubmit jobs?
When I didn't specify the registry explicitly in .prow.sh the envvar wasn't set in the gcr_cloud_build call.

I'm getting the following errors
```
#9 [auth] k8s-staging-sig-storage/csi-proxy:pull,push token for gcr.io
#9 DONE 0.0s

#8 exporting to image
#8 pushing layers 2.2s done
#8 ERROR: failed to authorize: failed to fetch oauth token: unexpected status: 403 Forbidden
------
 > exporting to image:
------
error: failed to solve: failed to fetch oauth token: unexpected status: 403 Forbidden
+ docker buildx rm multiarchimage-buildertest
make: *** [release-tools/build.make:149: push-multiarch-csi-proxy] Error 1
+ EXIT_VALUE=2
+ set +o xtrace
```

.cloudbuild.sh
```diff
@@ -1,11 +1,8 @@
 #! /bin/bash
 
-. release-tools/prow.sh
-
-# Extract tag-n-hash value from GIT_TAG (form vYYYYMMDD-tag-n-hash) for REV value.
-REV=v$(echo "$GIT_TAG" | cut -f3- -d 'v')
+: ${CSI_PROW_BUILD_PLATFORMS:="windows amd64 .exe nanoserver:1809; windows amd64 .exe nanoserver:ltsc2022"}
+: ${REGISTRY_NAME:="gcr.io/k8s-staging-sig-storage"}
```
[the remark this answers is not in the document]
mauriciopoppe: This might make the presubmit job pass but I think I might be forgetting about the post submit image push to `gcr.io/k8s-staging-sig-storage`, I'll check that too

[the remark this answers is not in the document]
mauriciopoppe:
```
: ${REGISTRY_NAME:="gcr.io/k8s-staging-sig-storage"}
```

shouldn't be set in this file, this is the file that runs on postsubmit

.prow.sh
```diff
@@ -4,7 +4,8 @@
 
 # # Only these tests make sense for csi-proxy
 : ${CSI_PROW_TESTS:="unit"}
-: ${CSI_PROW_BUILD_PLATFORMS:="windows amd64 .exe"}
+: ${CSI_PROW_BUILD_PLATFORMS:="windows amd64 .exe nanoserver:1809"}
+: ${REGISTRY_NAME:="gcr.io/k8s-staging-sig-storage"}
```
mauriciopoppe, line 8: This should be:

: ${REGISTRY_NAME:="gcr.io/$PROJECT"}

.prow.sh
```diff
@@ -4,7 +4,8 @@
 
 # # Only these tests make sense for csi-proxy
 : ${CSI_PROW_TESTS:="unit"}
-: ${CSI_PROW_BUILD_PLATFORMS:="windows amd64 .exe"}
+: ${CSI_PROW_BUILD_PLATFORMS:="windows amd64 .exe nanoserver:1809"}
```
mauriciopoppe, line 7: CSI_PROW_BUILD_PLATFORMS should already have a default, I think we should remove this line and use that default instead

.prow.sh
```diff
@@ -14,3 +15,6 @@ ensure_paths
 # main
 run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make all "GOFLAGS_VENDOR=${GOFLAGS_VENDOR}" "BUILD_PLATFORMS=${CSI_PROW_BUILD_PLATFORMS}"
 run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make -k test "GOFLAGS_VENDOR=${GOFLAGS_VENDOR}" 2>&1 | make_test_to_junit
+
+# build / push multi-arch images for validation
+gcr_cloud_build
```
mauriciopoppe, line 20: for clarification, this file .prow.sh is what runs in the presubmit jobs and .cloudbuild.sh is the one that runs on postsubmit jobs

mauriciopoppe:
@ddebroy, @jingxu97 and I talked about supporting this mode, my opinion is:

- having the binary running in a container makes its deployment easier than what it's now, CSI Drivers are in control of creating a bundle with CSI Proxy too and there's no need to do the node bootstrapping downloading the binary and running it as a Windows service.
- supporting a new artifact (the image) is harder, every time we create a new release we also have to promote the staging image to the k8s.gcr.io release project, the issue here is that if we push an image we have to support it for the foreseeable future.
- this PR is a step in between but might not be needed, HostProcess container is already beta in 1.23, I'd put more effort in modifying this repo to become a library and instead run the CSI Drivers/Addons with the HostProcess container setup.

marosset:
> @ddebroy, @jingxu97 and I talked about supporting this mode, my opinion is:
> 
> * having the binary running in a container makes its deployment easier than what it's now, CSI Drivers are in control of creating a bundle with CSI Proxy too and there's no need to do the node bootstrapping downloading the binary and running it as a Windows service.
> * supporting a new artifact (the image) is harder, every time we create a new release we also have to promote the staging image to the k8s.gcr.io release project, the issue here is that if we push an image we have to support it for the foreseeable future.
> * this PR is a step in between but might not be needed, HostProcess container is already beta in 1.23, I'd put more effort in modifying this repo to become a library and instead run the CSI Drivers/Addons with the HostProcess container setup.

I've been out of the office for a bit but am back now.
This makes sense. If we go the library route then we'd need to update many projects - correct?

mauriciopoppe:
> This makes sense. If we go the library route then we'd need to update many projects - correct?

Correct, we would:

- update this codebase so that the client library that it exposes doesn't use gRPC and instead just uses the server methods
- update CSI Drivers and addons to use CSI Proxy as a library

There's a doc that @ddebroy started but I haven't had a chance to complete yet: https://docs.google.com/document/d/15lgXyWytWSg1PuEcHl72eqgLoXc_i8kXmQcrXhRpjO0/edit#heading=h.6vn5bqze2ffs

marosset:
/close
I'm moving this work to github.com/kubernetes-sigs/sig-windows-tools with https://github.com/kubernetes-csi/csi-proxy/pull/179

PR #183, opened 2021-12-23
[WIP] Add the Lsdir API to the Filesystem v1 API Group

mauriciopoppe:
**What type of PR is this?**
/kind api-change
/kind feature

Adds the `Lsdir` API to list files in a directory through CSI Proxy, we're breaking the rule of not re-releasing CSI Proxy v1 but this is a new API so I think that the changes should be backwards compatible

```release-note
Added the Lsdir API to the Filesystem V1 API group to list files under a directory
```

/cc @jingxu97 @ddebroy


mauriciopoppe:
We abandoned this API in favor of #186 and #189

PR #188, opened 2022-01-07
Bump volume api from v1 to v2alpha1

mauriciopoppe:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
/kind feature

**What this PR does / why we need it**:
Bumps the v1 volume API to v2alpha1, we're planning on adding a new v2alpha1 API that finds the closes volume from a target path, the files were generated with this script:
```
API_GROUP=volume OLD_API_VERSION=v1 NEW_API_VERSION=v2alpha1 scripts/bump-version.sh
```

**Does this PR introduce a user-facing change?**:

```release-note
New API version v2alpha1 for the Volume API Group.
```

/cc @jingxu97 @ddebroy

PR #189, opened 2022-01-07
Add GetClosestVolumeIDFromTargetPath API to the Volume API Group

mauriciopoppe:
**What type of PR is this?**
>
/kind api-change
/kind feature

Adds the new Volume API, `GetClosestVolumeIDFromTargetPath`, it allows getting the closest volume from a target path handling more scenarios as described below.

The algorithm is:

- given a target path do the following for a max of 256 iterations (1)
  - if it's a symlink and has the form `Volume{` then its a volume, return it
  - if it's a symlink and points to a location that exists then follow it and go to 1
  - if it's not a symlink then move up one level in the hierarchy and go to 1
  - if after moving one level up we get to the same location then we're at the root, return the volume corresponding to the boot disk

In the following example all the calls to `GetClosestVolumeIDFromTargetPath` are with paths inside `/mnt/disks`:

```
/               (Volume{A}) # boot disk
/foo/bar
/mnt/volume/  (symlink -> Volume{B})
  shared-disk/
     foo/bar/shared0/
     shared1/
/mnt/single0/ (symlink -> Volume{C})  # standalone disk
/mnt/disks/ (discovery directory)
  shared0/ (symlink -> /mnt/volume/shared-disk/foo/bar/shared0) # returns Volume{B}
  shared1/ (symlink -> /mnt/volume/shared-disk/shared1) # returns Volume{B}
  single0/ (symlink -> /mnt/single0) # returns Volume{C}
  volume0/ (symlink -> Volume{D}) # returns Volume{D}
  boot0/   (symlink -> /foo/bar) # returns Volume{A}
```

```release-note
Add Volume API GetClosestVolumeIDFromTargetPath to find the closest Volume from a given path.
```

/hold 

Depends on https://github.com/kubernetes-csi/csi-proxy/pull/188

/cc @jingxu97 @ddebroy @msau42


pkg/os/volume/api.go
```diff
@@ -268,6 +272,96 @@ func getTarget(mount string) (string, error) {
...
+		}
+		isSymlink := fi.Mode()&os.ModeSymlink != 0
+
+		if isSymlink {
+			target, err := dereferenceSymlink(candidatePath)
+			if err != nil {
+				return "", err
+			}
+			// if it has the form Volume{volumeid}\ then it's a volume
+			if strings.HasPrefix(target, "Volume") {
```
ddebroy, line 307: I feel the Volume prefix check can be tightened with a regex to exactly match `Volume{...}` format. If a disk layout has symlink chains with links named `VolumeData1` and `VolumeData2` that are actually links to `Volume{A}` and `Volume{B}`, the `HasPrefix` logic would return incorrect results.

ddebroy (commented):
Recommend tightening the prefix check for Volume. I think the Volume prefix check in [k/k](https://github.com/kubernetes/kubernetes/blob/c1e69551be1a72f0f8db6778f20658199d3a686d/pkg/volume/util/subpath/subpath_windows.go#L62) should also be tightened with a regex.

mauriciopoppe:
/hold cancel

pull-kubernetes-csi-csi-proxy-integration passed, @jingxu97 it's now ready for review

pkg/os/volume/api.go
```diff
@@ -268,6 +277,96 @@ func getTarget(mount string) (string, error) {
...
+
+// findClosestVolume finds the closest volume id for a given target path
+// by following symlinks and moving up in the filesystem, if after moving up in the filesystem
+// we get to a DriveLetter then the volume corresponding to this drive letter is returned instead.
+func findClosestVolume(path string) (string, error) {
+	candidatePath := path
+
+	// run in a bounded loop to avoid doing an infinite loop
+	// while trying to follow symlinks
+	for i := 0; i < 256; i += 1 {
```
jingxu97, line 299: I think 256 is just an arbitrary number we pick, right?
if you have some references, pls add it here. 
Also add this to comment, we only follow up to 256?

pkg/os/volume/api.go
```diff
@@ -41,6 +45,11 @@ type VolumeAPI struct{}
 // verifies that the API is implemented
 var _ API = &VolumeAPI{}
 
+var (
+	// VolumeRegexp matches a Volume
+	VolumeRegexp = regexp.MustCompile(`^Volume\{[\w-]*\}`)
```
jingxu97, line 50: could you give some example of volumes?

jingxu97: it should ends with \\\}?

pkg/os/volume/api.go
```diff
@@ -268,6 +277,96 @@ func getTarget(mount string) (string, error) {
...
+	}
+
+	return "", fmt.Errorf("Failed to find the closest volume for path=%s", path)
+}
+
+// dereferenceSymlink dereferences the symlink `path` and returns the stdout.
+func dereferenceSymlink(path string) (string, error) {
+	cmd := exec.Command("powershell", "/c", fmt.Sprintf(`(Get-Item -Path %s).Target`, path))
+	klog.V(8).Infof("About to execute: %q", cmd.String())
+	targetb, err := cmd.Output()
```
jingxu97, line 344: not sure it is the case, sometimes the error is empty, but command actually return some error messages, so targetb will be some strings. In this case just want to make sure the logic can handle it.

pkg/os/volume/api.go
mauriciopoppe: Added comments about the why

pkg/os/volume/api.go
mauriciopoppe: added some examples

pkg/os/volume/api.go
mauriciopoppe: true, `cmd.Output` returns the combined output, places where this is called ensure that the path sent is a symlink however another way to test it would be by analyzing stderr apart from stdout.

The existing `GetVolumeIDFromTargetPath` doesn't do a separate analysis either, it's using the combined output and relying on err for errors

pkg/os/volume/api.go
```diff
@@ -268,6 +283,104 @@ func getTarget(mount string) (string, error) {
...
+
+		if isSymlink {
+			target, err := dereferenceSymlink(candidatePath)
+			if err != nil {
+				return "", err
+			}
+			// if it has the form Volume{volumeid} then it's a volume
+			if VolumeRegexp.Match([]byte(target)) {
+				// symlinks that are pointing to Volumes don't have this prefix
+				target = "\\\\?\\" + target
```
jingxu97, line 328: should we check prefix just in case before adding it?

pkg/os/volume/api.go
jingxu97: I just wondering whether it is possible to make sure it return a valid path or a volume id, instead of some long string

mauriciopoppe: I think that if the target doesn't exist it'll return an error that's propagated, so at this point we know that it's a path or a volumeId, for anything in stderr I'm not checking though, I'll only analyze the output of stdout and error if stderr has contents

PR #186, opened 2022-01-06
Add RmdirContents API in the Fileystem v2alpha1 API Group

mauriciopoppe:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
/kind api-change
/kind feature

**What this PR does / why we need it**:
Adds the new Filesystem API, `RmdirContents`, it removes the contents of a directory (but not the directory itself).

**Does this PR introduce a user-facing change?**:

```release-note
Add Filesystem API RmdirContents to remove the contents of a directory.
```

/cc @jingxu97 @ddebroy

PR #190, opened 2022-01-08
Scripts to run the integration tests with a real kubernetes VM

mauriciopoppe:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Some of our integration tests are run by us manually because we didn't have HyperV in a Windows VM, I recently added it to kubernetes so I'm planing on using a kubernetes cluster to run the integration tests, the algorithm is straightforward and is in `scripts/run-integration.sh`.

The next step is to add the config in test-infra to create a cluster, run the script and delete the cluster.

```release-note
NONE
```

/cc @jingxu97 @ddebroy


andyzhangx:
does test-infra supports creating windows VM? I think we could also run this integration test in github action which supports Windows node.

mauriciopoppe:
hey @andyzhangx, we're already running most of the integration tests with github actions however there are some tests that use VHDs (which require Hyper-V to be enabled in the VMs) and unfortunately the github actions VMs can't be configured to have this feature enabled, these tests are skipped in github actions.

In the main kubernetes repo it's possible to create a cluster with Windows VMs with kube-up.sh, recently I also added support to enable HyperV on https://github.com/kubernetes/kubernetes/pull/105999 which means that we can use a Windows GCE VM with HyperV enabled to run the integration tests through ssh 🙂.

mauriciopoppe:
@ddebroy @jingxu97 please review this PR again, the integration tests are passing in prow  🎉

scripts/utils.sh
```diff
@@ -0,0 +1,70 @@
...
+}
+
+compile_csi_proxy() {
+  echo "Compiling CSI Proxy"
+  make -C $pkgdir build
+}
+
+compile_csi_proxy_integration_tests() {
+  echo "Compiling CSI Proxy integration tests"
+  GOOS=windows GOARCH=amd64 go test -c $pkgdir/integrationtests -o $pkgdir/bin/integrationtests.test.exe
```
jingxu97, line 29: integrationtests.test.exe to integrationtests.exe?

scripts/run-integration.sh
```diff
@@ -0,0 +1,30 @@
...
+pkgdir=${GOPATH}/src/github.com/kubernetes-csi/csi-proxy
+source $pkgdir/scripts/utils.sh
+
+main() {
+  compile_csi_proxy
+  compile_csi_proxy_integration_tests
+  sync_csi_proxy
+  sync_csi_proxy_integration_tests
+  sync_powershell_utils
+  restart_csi_proxy
```
jingxu97, line 26: why restart? it was already started?

scripts/utils.sh
mauriciopoppe: this is similar to the e2e test binary, from `e2e.go` we generate the binary `e2e.test`, if I don't use the output flag it'll generate it with that name `integrationtests.test.exe`, I wrote `-o ` to send it to the `bin/` folder

scripts/run-integration.sh
mauriciopoppe: the idea is to test the latest binary and not the one that's preinstalled and running in the node, `sync_csi_proxy` only copies the binary to the host, `restart_csi_proxy` restarts it

mauriciopoppe:
Discussed with @jingxu97 offline and she's ok with these changes, @ddebroy could you please take a look? I'd really like to run the integration tests with #189 and #186

PR #196, opened 2022-01-27
skip resizing volume when extent is less than the minimum of 1MB

### LINKED ISSUE #195: Pod remains in the Pending state with error - Resize-Partition : Size Not Supported
**What happened**:
Created 2Gi PVC, PV is provisioned with 2Gi 
Extended PVC Capacity to 3Gi

```
$ kubectl describe pvc example-windows-pvc-1
Name:          example-windows-pvc-1
Namespace:     default
StorageClass:  example-windows-sc
Status:        Bound
Volume:        pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: csi.vsphere.vmware.com
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      2Gi
Access Modes:  RWO
VolumeMode:    Filesystem
Mounted By:    example-windows-pod-1
Conditions:
  Type                      Status  LastProbeTime                     LastTransitionTime                Reason  Message
  ----                      ------  -----------------                 ------------------                ------  -------
  FileSystemResizePending   True    Mon, 01 Jan 0001 00:00:00 +0000   Mon, 24 Jan 2022 11:17:06 -0800           Waiting for user to (re-)start a pod to finish file system resize of volume on node.
```

PV got extended to 3 Gi

```
$ kubectl describe pv pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67
Name:            pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67
Labels:          <none>
Annotations:     pv.kubernetes.io/provisioned-by: csi.vsphere.vmware.com
Finalizers:      [kubernetes.io/pv-protection external-attacher/csi-vsphere-vmware-com]
StorageClass:    example-windows-sc
Status:          Bound
Claim:           default/example-windows-pvc-1
Reclaim Policy:  Delete
Access Modes:    RWO
VolumeMode:      Filesystem
Capacity:        3Gi
Node Affinity:   <none>
Message:         
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            csi.vsphere.vmware.com
    VolumeHandle:      6d5caf4c-36ef-49f1-a804-0269c785b40d
    ReadOnly:          false
    VolumeAttributes:      storage.kubernetes.io/csiProvisionerIdentity=1643013955993-8081-csi.vsphere.vmware.com
                           type=vSphere CNS Block Volume
Events:                <none>
```

Created Pod using this volume
Pod remains in the Pending State with error - `Resize-Partition : Size Not Supported`.

```
Name:         example-windows-pod-1
Namespace:    default
Priority:     0
Node:         tkg-vc-antrea-md-0-windows-containerd-74d9795f7b-2kg72/10.92.213.208
Start Time:   Thu, 27 Jan 2022 09:09:59 -0800
Labels:       <none>
Annotations:  <none>
Status:       Pending
IP:           
Containers:
  test-container-1:
    Container ID:  
    Image:         mcr.microsoft.com/windows/servercore:ltsc2019
    Image ID:      
    Port:          <none>
    Host Port:     <none>
    Command:
      powershell.exe
      -Command
      while (1) { Add-Content -Encoding Ascii C:\test\data.txt $(Get-Date -Format u); sleep 1 }
    State:          Waiting
      Reason:       ContainerCreating
    Ready:          False
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /mnt/test-volume-1/ from test-volume-1 (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-2trft (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             False 
  ContainersReady   False 
  PodScheduled      True 
Volumes:
  test-volume-1:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  example-windows-pvc-1
    ReadOnly:   false
  kube-api-access-2trft:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   BestEffort
Node-Selectors:              kubernetes.io/hostname=tkg-vc-antrea-md-0-windows-containerd-74d9795f7b-2kg72
Tolerations:                 node.kubernetes.io/not-ready:NoExecute for 300s
                             node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type     Reason                  Age        From                                                             Message
  ----     ------                  ----       ----                                                             -------
  Normal   Scheduled               <unknown>                                                                   Successfully assigned default/example-windows-pod-1 to tkg-vc-antrea-md-0-windows-containerd-74d9795f7b-2kg72
  Normal   SuccessfulAttachVolume  41m        attachdetach-controller                                          AttachVolume.Attach succeeded for volume "pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67"
  Warning  FailedMount             40m        kubelet, tkg-vc-antrea-md-0-windows-containerd-74d9795f7b-2kg72  MountVolume.MountDevice failed while expanding volume for volume "pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67" : Expander.NodeExpand failed to expand the volume : rpc error: code = Internal desc = error when resizing filesystem on volume "6d5caf4c-36ef-49f1-a804-0269c785b40d" on node: error when resizing filesystem on devicePath \\?\Volume{3df70ab1-1cfe-4047-bd7f-a5d6e44b754e}\ and volumePath \var\lib\kubelet\plugins\kubernetes.io\csi\pv\pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67\globalmount, err: rpc error: code = Unknown desc = error resizing volume. cmd: Get-Volume -UniqueId "\\?\Volume{3df70ab1-1cfe-4047-bd7f-a5d6e44b754e}\" | Get-Partition | Resize-Partition -Size 3204431360, output: Resize-Partition : Size Not Supported
```

> error when resizing filesystem on volume "6d5caf4c-36ef-49f1-a804-0269c785b40d" on node: error when resizing filesystem on devicePath \\?\Volume{3df70ab1-1cfe-4047-bd7f-a5d6e44b754e}\ and volumePath \var\lib\kubelet\plugins\kubernetes.io\csi\pv\pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67\globalmount, err: rpc error: code = Unknown desc = error resizing volume. cmd: Get-Volume -UniqueId "\\?\Volume{3df70ab1-1cfe-4047-bd7f-a5d6e44b754e}\" | Get-Partition | Resize-Partition -Size 3204431360, output: Resize-Partition : Size Not Supported

Pod should not remain in this pending state

**How to reproduce it**:
extend volume size on the new volume before using the volume with the pod.

**Environment**:
- CSI Driver version: vSphere CSI Driver 2.4.1
- Kubernetes version (use `kubectl version`): v1.22.3
- OS (e.g. from /etc/os-release):  Windows Server 2019, Version 1809
- Kernel (e.g. `uname -a`): 
- Install tools: CSI proxy v1.0.2
- Others:

divyenpatel:
**What type of PR is this?**
/kind bug

This PR is skipping resize of the volume when extent is less than the minimum of 1MB

Fixes #195 

Validated changes with vSphere CSI Driver. Volume is getting expanded when the volume is resized before creating a filesystem on it.

```release-note
skip resizing volume when extent is less than the minimum of 1MB
```

### DIFF
--- a/pkg/os/volume/api.go
+++ b/pkg/os/volume/api.go
@@ -11,6 +11,8 @@ import (
 	"strconv"
 	"strings"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/klog/v2"
 )
 
@@ -56,6 +58,10 @@ var (
 	// \\?\Volume{452e318a-5cde-421e-9831-b9853c521012}\
 	VolumeRegexp = regexp.MustCompile(`Volume\{[\w-]*\}`)
 )
+
+// MinimumExpandSize is the minimum size required to
+// expanding the partition
+const MinimumExpandSize = 1024 * 1024
 
 // New - Construct a new Volume API Implementation.
 func New() VolumeAPI {
@@ -186,18 +192,24 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
 	} else {
 		finalSize = size
 	}
+	klog.V(4).Infof("finalSize for resize of the volume: %s is %d ", volumeID, finalSize)
 
 	currentSize, err := getVolumeSize(volumeID)
 	if err != nil {
 		return fmt.Errorf("error getting the current size of volume (%s) with error (%v)", volumeID, err)
 	}
-
+	klog.V(4).Infof("currentSize of the volume: %s is %d ", volumeID, currentSize)
 	//if the partition's size is already the size we want this is a noop, just return
 	if currentSize >= finalSize {
 		klog.V(2).Infof("Attempted to resize volume %s to a lower size, from currentBytes=%d wantedBytes=%d", volumeID, currentSize, finalSize)
 		return nil
 	}
-
+	// Resize-Partition requires minimum extent size 1 MB
+	// if extent size for resize partition is less than 1 MB, skip expanding volume
+	if finalSize-currentSize < MinimumExpandSize {
+		return status.Errorf(codes.OutOfRange,
+			"minimum extent size is 1 MB. Skip resize for volume %s from currentBytes=%d to wantedBytes=%d ", volumeID, currentSize, finalSize)
+	}
 	cmd = fmt.Sprintf("Get-Volume -UniqueId \"%s\" | Get-Partition | Resize-Partition -Size %d", volumeID, finalSize)
 	out, err = runExec(cmd)
 	if err != nil {
--- a/integrationtests/volume_v2alpha1_test.go
+++ b/integrationtests/volume_v2alpha1_test.go
@@ -13,6 +13,8 @@ import (
 	"github.com/kubernetes-csi/csi-proxy/client/api/volume/v2alpha1"
 	diskv1client "github.com/kubernetes-csi/csi-proxy/client/groups/disk/v1"
 	v2alpha1client "github.com/kubernetes-csi/csi-proxy/client/groups/volume/v2alpha1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // volumeInit initializes a volume, it creates a VHD, initializes it,
@@ -253,6 +255,30 @@ func v2alpha1MountVolumeTests(diskClient *diskv1client.Client, volumeClient *v2a
 	// resizing from 1GB to approximately 1.5GB
 	if !sizeIsAround(t, volumeStatsResponse.TotalBytes, newVolumeSize) {
 		t.Fatalf("VolumeSize reported should be greater than the old size, it is %v", volumeStatsResponse.TotalBytes)
+	}
+
+	// extending volume less than 10 MB should be failed
+	resizeVolumeRequest = &v2alpha1.ResizeVolumeRequest{
+		VolumeId: volumeID,
+		// set new size less than 10 MB
+		SizeBytes: newVolumeSize + (1024 * 512),
+	}
+	_, err = volumeClient.ResizeVolume(context.TODO(), resizeVolumeRequest)
+	if err == nil {
+		t.Fatal("Volume resize request should be failed")
+	}
+	status, ok := status.FromError(err)
+	if !ok || status.Code() != codes.OutOfRange {
+		t.Fatal("Volume resize request should be failed with OutOfRange")
+	}
+	// Verify volume size after expected failed resize attempt
+	volumeStatsResponseAfterResize, err := volumeClient.GetVolumeStats(context.TODO(), volumeStatsRequest)
+	if err != nil {
+		t.Fatalf("VolumeStats request after resize error: %v", err)
+	}
+	if volumeStatsResponseAfterResize.TotalBytes != volumeStatsResponse.TotalBytes {
+		t.Fatalf("VolumeSize should not be extended from %d to %d", volumeStatsResponse.TotalBytes,
+			volumeStatsResponseAfterResize.TotalBytes)
 	}
 
 	volumeDiskNumberRequest := &v2alpha1.GetDiskNumberFromVolumeIDRequest{


pkg/os/volume/api.go
```diff
@@ -170,18 +170,22 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
 	} else {
 		finalSize = size
 	}
+	klog.V(2).Infof("finalSize for resize of the volume: %s is %d ", volumeID, finalSize)
```
mauriciopoppe, line 173: V(2) might be too verbose, we don't use `-v` in OSS though

mauriciopoppe (commented):
It'd be nice to have an integration test too, the code is in integrationtests/volume_v2alpha1_test.go

pkg/os/volume/api.go
```diff
@@ -170,18 +170,22 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
...
 	}
-
+	klog.V(2).Infof("currentSize of the volume: %s is %d ", volumeID, currentSize)
 	//if the partition's size is already the size we want this is a noop, just return
 	if currentSize >= finalSize {
 		klog.V(2).Infof("Attempted to resize volume %s to a lower size, from currentBytes=%d wantedBytes=%d", volumeID, currentSize, finalSize)
 		return nil
 	}
-
+	if finalSize-currentSize < 1048576 {
```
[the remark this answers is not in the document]
xing-yang: Can you add some explanation here to say that this could happen if the volume is already resized?

mauriciopoppe:
pull-kubernetes-csi-csi-proxy-integration failed because of a pending merge of #186 or #189

pkg/os/volume/api.go
```diff
@@ -41,6 +41,8 @@ type VolumeAPI struct{}
 // verifies that the API is implemented
 var _ API = &VolumeAPI{}
 
+const minimum_extent_size = 1024 * 1024
```
mauriciopoppe, line 44: Thanks for adding an integration tests, sorry to be annoying but this should be `MinimumExtentSize`, also, is `expand` a better word for it?

pkg/os/volume/api.go
divyenpatel: fixed. `used klog.V(4)`.

pkg/os/volume/api.go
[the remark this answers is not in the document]
divyenpatel: added comment

pkg/os/volume/api.go
divyenpatel: done.

jingxu97:
just want to understand when this scenario might happen. In your test case, you increase from 2GB to 3GB, how it caused failure?

divyenpatel:
> just want to understand when this scenario might happen. In your test case, you increase from 2GB to 3GB, how it caused failure?

We created 2 GB volume and then expanded the volume size to 3 GB, without creating the filesystem on the volume.
Then we created a Pod. While creating pod, the volume got formatted and the filesystem is created with 3 GB ( around 3204382720 bytes), later when the volume is attempted to expand, the final size to expand the volume is determined as 3204431360 bytes, which is less than the minimum of 1 MB requirement for resize-parition.

This issue is not happening for the case, when we create a 2 GB volume, create a pod using this volume, then we expand the disk to 3 GB and re-create the pod to expand the filesystem.

pkg/os/volume/api.go
```diff
@@ -170,18 +174,24 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
...
+	klog.V(4).Infof("currentSize of the volume: %s is %d ", volumeID, currentSize)
 	//if the partition's size is already the size we want this is a noop, just return
 	if currentSize >= finalSize {
 		klog.V(2).Infof("Attempted to resize volume %s to a lower size, from currentBytes=%d wantedBytes=%d", volumeID, currentSize, finalSize)
 		return nil
 	}
-
+	// Resize-Partition requires minimum extent size 1 MB
+	// if extent size for resize partition is less than 1 MB, skip expanding volume
+	if finalSize-currentSize < MinimumExpandSize {
```
mauriciopoppe, line 191: While the error is handled by CSI Proxy it'd be great to let the user know what happened and that can only happen if CSI Proxy throws the error and the CSI Drivers handles it.

divyenpatel: do you mean return an error with the code using https://github.com/grpc/grpc-go/blob/master/status/status.go#L62-L65 and handle this error in the driver to make a decision about failure or success?

```goalng
	if finalSize-currentSize < MinimumExpandSize {
		return status.Errorf(codes.OutOfRange, 
			"minimum extent size is 1 MB. Skip resize for volume %s from currentBytes=%d to wantedBytes=%d ", volumeID, currentSize, finalSize)
	}
```

mauriciopoppe: `codes.OutOfRange` is a good idea, yes I think that the driver should handle the error, if no resize happened then it can log that too without erroring for better visibility rather than checking the CSI Proxy logs

jingxu97:
> later when the volum

I see, interesting scenario. In your case, when volume is 2GB, no pod is using it, so no filesystem is created. Now when you resize to 3G, at this moment, if no pod is created, I assume no filesystem is created either, right? Or in this case, file system is created even without pod using it?

divyenpatel:
> > later when the volum
> 
> I see, interesting scenario. In your case, when volume is 2GB, no pod is using it, so no filesystem is created. Now when you resize to 3G, at this moment, if no pod is created, I assume no filesystem is created either, right? Or in this case, file system is created even without pod using it?

when I resize the volume to 3 GB, no file system was present on the volume.
later when I created a Pod using extended volume, Pod got stuck at `MountVolume.MountDevice failed while expanding volume for volume "pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67" : Expander.NodeExpand failed `

jingxu97:
> > > later when the volum
> > 
> > 
> > I see, interesting scenario. In your case, when volume is 2GB, no pod is using it, so no filesystem is created. Now when you resize to 3G, at this moment, if no pod is created, I assume no filesystem is created either, right? Or in this case, file system is created even without pod using it?
> 
> when I resize the volume to 3 GB, no file system was present on the volume. later when I created a Pod using extended volume, Pod got stuck at `MountVolume.MountDevice failed while expanding volume for volume "pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67" : Expander.NodeExpand failed `

 
The change here is good. In addition, I think some logic in volume expansion should handle better this case to avoid reformatting it since it was already formatted with the correct file size.

pkg/os/volume/api.go
divyenpatel: addressed review comment with this commit - https://github.com/kubernetes-csi/csi-proxy/pull/196/commits/603b06b028cdc2b1dc9e1c657d4294b2d8d59df4

divyenpatel:
@jingxu97 @mauriciopoppe I have squashed all commits. Can we merge this PR?

mauriciopoppe:
it looks good to me, @jingxu97 will lgtm and approve

jingxu97:
@divyenpatel this failure only happen in Windows case, right?

jingxu97:
> > just want to understand when this scenario might happen. In your test case, you increase from 2GB to 3GB, how it caused failure?
> 
> We created 2 GB volume and then expanded the volume size to 3 GB, without creating the filesystem on the volume. Then we created a Pod. While creating pod, the volume got formatted and the filesystem is created with 3 GB ( around 3204382720 bytes), later when the volume is attempted to expand, the final size to expand the volume is determined as 3204431360 bytes, which is less than the minimum of 1 MB requirement for resize-parition.
> 
> This issue is not happening for the case, when we create a 2 GB volume, create a pod using this volume, then we expand the disk to 3 GB and re-create the pod to expand the filesystem.

cc @gnufied Do you think resize controller can avoid triggering resize in this scenario? When volume is provisioned with 2GB, it is not formatted. Then it gets resized to 3GB and gets formatted. Then in this case, no resize of file system is needed.

jingxu97:
> @divyenpatel this failure only happen in Windows case, right?

If we return error in csi-proxy api, then in current logic, all the driver might need to modify to handle this case, otherwise it will fail here.

gnufied:
>when I resize the volume to 3 GB, no file system was present on the volume.
later when I created a Pod using extended volume, Pod got stuck at MountVolume.MountDevice failed while expanding volume for volume "pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67" : Expander.NodeExpand failed

Isn't `NodeExpand` supposed to be idempotent and hence if volume was already of size 3GB and expansion on node is again called for 3GB,the driver should return success and everything should fall in place.

jingxu97:
> > when I resize the volume to 3 GB, no file system was present on the volume.
> > later when I created a Pod using extended volume, Pod got stuck at MountVolume.MountDevice failed while expanding volume for volume "pvc-021b5d9c-ba3a-4c9e-ae22-df2cbfce9e67" : Expander.NodeExpand failed
> 
> Isn't `NodeExpand` supposed to be idempotent and hence if volume was already of size 3GB and expansion on node is again called for 3GB,the driver should return success and everything should fall in place.

I might be wrong, in this case, nodeexpand is only called once. The volume is formatted with 3GB, but in windows, actual file system size will be less then 3GB. Then it calls nodeExpand trying to expand to 3GB and hit the error

gnufied:
> I might be wrong, in this case, nodeexpand is only called once. The volume is first formatted with 3GB, but in windows, actual file system size will be less then 3GB. Then it calls nodeExpand trying to expand to 3GB and hit the error.

It sounds like a driver bug. If file system was on 3GB disk (i.e I am not saying file system size but rather total disk space on which file system exists) and kubelet asked to expand to 3GB again, then driver should return success. Why does the driver returns error? Given current behaviour - `NodeExpand` is *never* going to be idempotent, because file system size is always going to be smaller than disk size.

pkg/os/volume/api.go
```diff
@@ -186,18 +192,24 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
...
+		return status.Errorf(codes.OutOfRange,
+			"minimum extent size is 1 MB. Skip resize for volume %s from currentBytes=%d to wantedBytes=%d ", volumeID, currentSize, finalSize)
+	}
```
gnufied, line 212: Does this fixes the issue we observed btw? It sounds like - this will prevent node-expansion if increased size is less than 1MB?

jingxu97:
> > I might be wrong, in this case, nodeexpand is only called once. The volume is first formatted with 3GB, but in windows, actual file system size will be less then 3GB. Then it calls nodeExpand trying to expand to 3GB and hit the error.
> 
> It sounds like a driver bug. If file system was on 3GB disk (i.e I am not saying file system size but rather total disk space on which file system exists) and kubelet asked to expand to 3GB again, then driver should return success. Why does the driver returns error? Given current behaviour - `NodeExpand` is _never_ going to be idempotent, because file system size is always going to be smaller than disk size.

I checked some code here, so during FormatVolume it uses the following command, no size needs to be specifies during format. System should try to format the disk as much as it can after reserving some space for system use. With 3GB disk, the file system size seems 3204382720bytes
```
Get-Volume -UniqueId \"%s\" | Format-Volume -FileSystem ntfs -Confirm:$false", volumeID
```
When driver calls ResizeVolume, it does not pass a size either which means resize to the max value that this disk can. So maxSize is calculated with following command. For 3GB disk, it returns 3204431360bytes
```
Get-Volume -UniqueId \"%s\" | Get-partition | Get-PartitionSupportedSize | Select SizeMax | ConvertTo-Json", volumeID
```
The code also get the current file system size with following command, the value is 3204382720bytes
```
(Get-Volume -UniqueId \"%s\" | Get-partition).Size", volumeID
```
If maxSize is the same as current size, the current logic will skip resize and return success. In this case it happens that maxSize is a little bit larger than current size, so it continues to run resize command and failed. 

I am wondering how linux handle this case, when tries to resize the volume which is already formatted under the same disk capacity using command like "resize2fs" for ext file system.

Two things I am thinking
1. whether we can avoid calling resize if format is already for the same disk capacity
2. in this resize is called for the same disk capacity that file system is formatted, how to identify this situation and return success instead of error. For example, when resize is called without passing a resize number, that means to use the max value disk can support, if it is very close to the existing value like this case. If it does pass a value, which means user explicitly want to resize this much, it can return error. (is this case ever used in any driver?)

divyenpatel:
@jingxu97 I have checked 4 CSI drivers, all of them call resize with  SizeBytes 0 to use the maximum available size. `&volume.ResizeVolumeRequest{VolumeId: devicePath, SizeBytes: 0}`


https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/pkg/mounter/safe_mounter_windows.go#L329
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/resizefs/resizefs_windows.go#L70-L72
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/mounter/safe_mounter_windows.go#L311
https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/pkg/csi/service/mounter/mounter_windows.go#L406

so this is affect all these drivers.

so to be on the safe side do you suggest checking if supplied SizeBytes is 0, and we find `if finalSize-currentSize < MinimumExpandSize {` then do not error out, but error out only when supplied SizeBytes is not zero and `if finalSize-currentSize < MinimumExpandSize`

gnufied:
>When driver calls ResizeVolume, it does not pass a size either which means resize to the max value that this disk can. So maxSize is calculated with following command. For 3GB disk, it returns 3204431360bytes

Why doesn't `ResizeVolume` get a size here? Kubelet always passes size to the CSI driver btw. 

> The code also get the current file system size with following command. If maxSize is the same as current size, the current logic will skip resize and return success. In this case it happens that maxSize is a little bit larger than current size, so it continues to run resize command and failed.

I wonder if there is a way to get disk size on which file system is, vs file system size. In Linux it is possible to get disk size on which file system exists vs file system size. That is one way to determine if file system on the disk needs expansion. This is similar to problem we ran into when a snapshot can be restored to size larger than original volume and hence we had to calculate disk space on which file system exits (which is *very* different from file system size).

jingxu97:
> resize2fs

We checked a few drivers, when calling resize, none of them pass a size value. I think it tries to resize to the max size that disk can possible hold.

I tried command resize2fs on a disk with file system already exist, it returns

```
resize2fs 1.46.2 (28-Feb-2021)
The filesystem is already 1572864 (4k) blocks long.  Nothing to do!
```

jingxu97:
> @jingxu97 I have checked 4 CSI drivers, all of them call resize with SizeBytes 0 to use the maximum available size. `&volume.ResizeVolumeRequest{VolumeId: devicePath, SizeBytes: 0}`
> 
> https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/pkg/mounter/safe_mounter_windows.go#L329 https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/resizefs/resizefs_windows.go#L70-L72 https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/mounter/safe_mounter_windows.go#L311 https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/pkg/csi/service/mounter/mounter_windows.go#L406
> 
> so this is affect all these drivers.
> 
> so to be on the safe side do you suggest checking if supplied SizeBytes is 0, and we find `if finalSize-currentSize < MinimumExpandSize {` then do not error out, but error out only when supplied SizeBytes is not zero and `if finalSize-currentSize < MinimumExpandSize`

I agree with the logic. Seems like linux can handle this well without extra logic to check, but Windows we have to handle it ourselves.

jingxu97:
@divyenpatel just to check the status of this PR?

jingxu97:
do we still need to merge this change?

jingxu97:
I don't have strong objection, @gnufied WDYT?
but the PR need to solve merge conflict. @divyenpatel

mauriciopoppe:
Development of the client/server model has moved to the v1.x branch, please change the base branch to v1.x

PR #210, opened 2022-06-28
Try to fix issue with 'Multiple connections to a server'

vitaliy-leschenko:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**
This PR allow to use pv/pvc for the same path more that once.
This PR allow to use many PVC based on storage class.

**Which issue(s) this PR fixes:**
Fixes https://github.com/kubernetes-csi/csi-driver-smb/issues/385

**Special notes for reviewer:**
This PR has known issue: it still doesn't work for many PVC with the same smb-path but with different credentials.

```release-note
Fix issue with impossible using a smb share more than once.
```

### DIFF
--- a/pkg/server/smb/server.go
+++ b/pkg/server/smb/server.go
@@ -65,7 +65,7 @@ func (s *Server) NewSmbGlobalMapping(context context.Context, request *internal.
 
 	mappingPath, err := getRootMappingPath(remotePath)
 	if err != nil {
-		return response, err;
+		return response, err
 	}
 
 	isMapped, err := s.hostAPI.IsSmbMapped(mappingPath)
@@ -133,7 +133,7 @@ func (s *Server) RemoveSmbGlobalMapping(context context.Context, request *intern
 
 	mappingPath, err := getRootMappingPath(remotePath)
 	if err != nil {
-		return response, err;
+		return response, err
 	}
 
 	err = s.hostAPI.RemoveSmbGlobalMapping(mappingPath)
--- a/pkg/server/smb/server.go
+++ b/pkg/server/smb/server.go
@@ -25,7 +25,7 @@ func normalizeWindowsPath(path string) string {
 	return normalizedPath
 }
 
-func normalizeMappingPath(path string) string {
+func getRootMappingPath(path string) (string, error) {
 	items := strings.Split(path, "\\")
 	parts := []string{}
 	for _, s := range items {
@@ -36,7 +36,13 @@ func normalizeMappingPath(path string) string {
 			}
 		}
 	}
-	return strings.ToLower("\\\\" + parts[0] + "\\" + parts[1])
+	if len(parts) != 2 {
+		klog.Errorf("remote path (%s) is invalid", path)
+		return nil, fmt.Errorf("remote path (%s) is invalid", path)
+	}
+	// parts[0] is a smb host name
+	// parts[1] is a smb share name
+	return strings.ToLower("\\\\" + parts[0] + "\\" + parts[1]), nil
 }
 
 func NewServer(hostAPI smb.API, fsServer *fsserver.Server) (*Server, error) {
@@ -57,7 +63,9 @@ func (s *Server) NewSmbGlobalMapping(context context.Context, request *internal.
 		return response, fmt.Errorf("remote path is empty")
 	}
 
-	mappingPath := normalizeMappingPath(remotePath)
+	if mappingPath, err := getRootMappingPath(remotePath); err != nil {
+		return response, err;
+	}
 
 	isMapped, err := s.hostAPI.IsSmbMapped(mappingPath)
 	if err != nil {
@@ -122,7 +130,10 @@ func (s *Server) RemoveSmbGlobalMapping(context context.Context, request *intern
 		return response, fmt.Errorf("remote path is empty")
 	}
 
-	mappingPath := normalizeMappingPath(remotePath)
+	if mappingPath, err := getRootMappingPath(remotePath); err != nil {
+		return response, err;
+	}
+
 	err := s.hostAPI.RemoveSmbGlobalMapping(mappingPath)
 	if err != nil {
 		klog.Errorf("failed RemoveSmbGlobalMapping %v", err)
--- a/pkg/server/smb/server.go
+++ b/pkg/server/smb/server.go
@@ -25,6 +25,20 @@ func normalizeWindowsPath(path string) string {
 	return normalizedPath
 }
 
+func normalizeMappingPath(path string) string {
+	items := strings.Split(path, "\\")
+	parts := []string{}
+	for _, s := range items {
+		if len(s) > 0 {
+			parts = append(parts, s)
+			if len(parts) == 2 {
+				break
+			}
+		}
+	}
+	return strings.ToLower("\\\\" + parts[0] + "\\" + parts[1])
+}
+
 func NewServer(hostAPI smb.API, fsServer *fsserver.Server) (*Server, error) {
 	return &Server{
 		hostAPI:  hostAPI,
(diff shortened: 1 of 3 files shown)


andyzhangx:
@vitaliy-leschenko thanks for the contribution. Write a file to count the smb mapping is not a good idea.

Current csi-proxy already supports PVs with same smb remote path since there is a `hostAPI.IsSmbMapped` check:
https://github.com/kubernetes-csi/csi-proxy/blob/1e9055abd899845e95a039772ff867636a3f2a0a/pkg/server/smb/server.go#L46-L51, so https://github.com/kubernetes-csi/csi-driver-smb/issues/385 should already be fixed by latest csi-proxy version.

The original problem is when there are two smb remote path with same smb server address and different share path, it would fail, below are details, e.g.
 - `\\20.xx.xx.56\share1`
 - `\\20.xx.xx.56\share2`

https://github.com/kubernetes-csi/csi-driver-smb/issues/219#issuecomment-781952587

vitaliy-leschenko:
@andyzhangx `s.hostAPI.IsSmbMapped(remotePath)` doesn't fix the issue because it doesn't work when we use a `storageClass`

vitaliy-leschenko:
Current solution (without this PR) doesn't work in this case:
1. one PV/PVC to \\smbserver\sharename\subpath1
2. one PV/PVC to \\smbserver\sharename\subpath2
Another case when the current solution fails:
two (or more) tries to use storageClass to \\smbserver\sharename because it generates pv/pvc
1. \\smbserver\sharename\pvc-%guid1%
2. \\smbserver\sharename\pvc-%guid2%

This PR a little bit change logic of using `*-SmbGlobalMapping` comandlets. When we need to create a new one, we create it for  `\\smbserver\sharename` instead of full path. It resolves the issue. But it creates another one: we need count global mappings because now we have one-to-many relation between pvc and global mapping. This PR tries to resolve this issue using special files. I understand that it isn't a brilliant solution but currently we don't have a better one.

@andyzhangx what do you think about it?

vitaliy-leschenko:
Currently no one calls `Remove-SmbGlobalMapping`. Another PR (https://github.com/kubernetes-csi/csi-driver-smb/pull/505) fixes it.

pkg/server/smb/server.go
```diff
@@ -25,6 +29,71 @@ func normalizeWindowsPath(path string) string {
...
+			}
+		}
+	}
+	return strings.ToLower("\\\\" + parts[0] + "\\" + parts[1])
+}
+
+func getMappingPathCount(mappingPath string) int {
+	klog.V(2).Infof("calling getMappingPathCount with mapping path %q", mappingPath)
+
+	path := basePath + "/" + strings.Replace(strings.TrimPrefix(mappingPath, "\\\\"), "\\", "/", -1)
```
andyzhangx, line 49: use such way to get <remotePath, localPath> mapping count is quite weird. not sure whether there is better way to achieve this.  cc @jingxu97 @ddebroy

vitaliy-leschenko: Has somebody seen this?

ddebroy:
Will it be possible for the CSI plugin to maintain the state introduced in this PR? csi-proxy should just be a thin wrapper over the OS API calls?

vitaliy-leschenko:
> Will it be possible for the CSI plugin to maintain the state introduced in this PR? csi-proxy should just be a thin wrapper over the OS API calls.

I think it possible. You are right, I forgot that `csi-proxy` it is just a proxy as temp solution.
I will do it.

vitaliy-leschenko:
@ddebroy @andyzhangx I moved logic with reference counting to csi-driver-smb PR.
Now it compatible with old versions of csi-driver-smb and contains only API calls.

pkg/server/smb/server.go
```diff
@@ -25,6 +25,20 @@ func normalizeWindowsPath(path string) string {
 	return normalizedPath
 }
 
+func normalizeMappingPath(path string) string {
+	items := strings.Split(path, "\\")
+	parts := []string{}
+	for _, s := range items {
+		if len(s) > 0 {
+			parts = append(parts, s)
+			if len(parts) == 2 {
```
andyzhangx, line 34: is `normalizeMappingPath` necessary here? I remember csi-proxy has already convert `/` to `\\` in somewhere.
this func assumes path only has two `\\`, it's not always that case.

vitaliy-leschenko: `normalizeMappingPath` gets from a full smb path the smb share name that we will use for mapping.
For example:
```
\\hostname\share\a\b\c\d  => \\hostname\share
\\hostname\share => \\hostname\share
```

csi-driver-smb has the same function for reference counting. Unfortunately, we can't send normalized value to csi-proxy because after create smb global mapping it also makes link to full remote path. I tried to avoid change of contract between csi-driver-smb and csi-proxy. So, this function exists in both repos.

andyzhangx: while if you change `\\hostname\share\a\b\c\d` to `\\hostname`, that already changes the mounting source, e.g. if user wants to access file under `\\hostname\share\a\b\c\d`, it's in the root dir, while with this PR, it's under `a\b\c\d` after mount.

vitaliy-leschenko: No.
When user want to mount `\\hostname\share\subpath` now we call:
```
New-SmbGlobalMapping -RemotePath "\\hostname\share\subpathA" -Credential $creds
New-Item -ItemType SymbolicLink "...\globalmount" -Target "\\hostname\share\subpathA"
```
This is a problem because we can't do:
```
New-SmbGlobalMapping -RemotePath "\\hostname\share\subpathB" -Credential $creds
New-Item -ItemType SymbolicLink "...\globalmount" -Target "\\hostname\share\subpathB"
```

This PR do:
```
New-SmbGlobalMapping -RemotePath "\\hostname\share" -Credential $creds
New-Item -ItemType SymbolicLink "...\globalmount" -Target "\\hostname\share\subpathA"
```
Then
```
New-Item -ItemType SymbolicLink "...\globalmount" -Target "\\hostname\share\subpathB"
```
The last mount doesn't call `New-SmbGlobalMapping` beause it already mounted.
And it works.

vitaliy-leschenko: So, user will have result as expected.

vitaliy-leschenko: subpath might be empty

andyzhangx: that makes sense, thanks for the explanation.

pkg/server/smb/server.go
```diff
@@ -25,6 +25,20 @@ func normalizeWindowsPath(path string) string {
 	return normalizedPath
 }
 
+func normalizeMappingPath(path string) string {
```
andyzhangx, line 28: `getRootMappingPath` is better

pkg/server/smb/server.go
```diff
@@ -25,6 +25,20 @@ func normalizeWindowsPath(path string) string {
...
+	parts := []string{}
+	for _, s := range items {
+		if len(s) > 0 {
+			parts = append(parts, s)
+			if len(parts) == 2 {
+				break
+			}
+		}
+	}
+	return strings.ToLower("\\\\" + parts[0] + "\\" + parts[1])
```
andyzhangx, line 39: if len(parts) < 2 it would panic

pkg/server/smb/server.go
```diff
@@ -25,6 +25,20 @@ func normalizeWindowsPath(path string) string {
 	return normalizedPath
 }
 
+func normalizeMappingPath(path string) string {
+	items := strings.Split(path, "\\")
```
andyzhangx, line 29: is it possible that `path` is like `//hostname/share` ? do we need to handle such case?

vitaliy-leschenko: I will check

pkg/server/smb/server.go
vitaliy-leschenko: I will check it and cover in unit tests

vitaliy-leschenko: done

pkg/server/smb/server.go
vitaliy-leschenko: done

pkg/server/smb/server.go
```diff
@@ -101,11 +131,17 @@ func (s *Server) RemoveSmbGlobalMapping(context context.Context, request *intern
 		return response, fmt.Errorf("remote path is empty")
 	}
 
-	err := s.hostAPI.RemoveSmbGlobalMapping(remotePath)
+	mappingPath, err := getRootMappingPath(remotePath)
```
mauriciopoppe, line 134: does the value of `mappingPath` change after the call on line 66?

vitaliy-leschenko: lines 66 and 134 relate to different methods

mauriciopoppe: Got it, thanks

vitaliy-leschenko:
@andyzhangx I've fixed tests. Please re-review.

andyzhangx (approved):
/lgtm
 I will leave @mauriciopoppe to approve, thanks for the contribution.

mauriciopoppe:
Interesting, someone approved it, it'd have been great to have this change as a single commit instead of 10 commits

shurick81:
@mauriciopoppe @andyzhangx do you by any chance know when to expect this release (1.1.2) in AKS? Is there a way to see in AKS which version of csi-proxy is deployed?

mauriciopoppe:
> Is there a way to see in AKS which version of csi-proxy is deployed?

I'm not familiar with AKS, however if you can SSH to a node you could find the version in the first lines of `C:\etc\kubernetes\logs\csi-proxy.log`, also, it might be possible that the path is different too. The kubernetes-csi community doesn't control the deployment of the binary, that's the task of the cluster administrator (in your case the devs that deploy Kubernetes in AKS).

PR #221, opened 2022-10-05
Migrate to v2.0 to become a library

### LINKED ISSUE #217: Refactor CSI Proxy to become a go library in favor of the client/server model
/assign @mauriciopoppe

alexander-ding:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

In light of HostProcess containers on K8s Windows, we no longer need the client/server model, with a separate binary running on each Windows node. Instead, we can provide CSI Proxy's functionalities as a Go library to be imported by CSI driver implementations running in HostProcess containers. The change involves a complete restructure of the package structure and bumping the Go version to v2. Each API group is now provided under /pkg/<API Group>, with the OS package written in /pkg/<API Group>/api. 

Part of #217 

```release-note
TODO
```


pkg/disk/disk.go
```diff
@@ -0,0 +1,238 @@
+package disk
```
mauriciopoppe, line 1: was this file copied or moved? I think we should move files first tracking them with git and then make changes

pkg/system/types.go
```diff
@@ -31,24 +31,24 @@ type StopServiceResponse struct {
...
 const (
-	SERVICE_STATUS_UNKNOWN          = 0
-	SERVICE_STATUS_STOPPED          = 1
-	SERVICE_STATUS_START_PENDING    = 2
-	SERVICE_STATUS_STOP_PENDING     = 3
-	SERVICE_STATUS_RUNNING          = 4
-	SERVICE_STATUS_CONTINUE_PENDING = 5
-	SERVICE_STATUS_PAUSE_PENDING    = 6
-	SERVICE_STATUS_PAUSED           = 7
+	SERVICE_STATUS_UNKNOWN ServiceStatus = iota
```
mauriciopoppe, line 34: cool, thanks for this change

integrationtests/system_test.go
```diff
@@ -67,36 +64,34 @@ func TestServiceCommands(t *testing.T) {
 		require.NoError(t, err, "failed unmarshalling json out=%v", out)
 
 		assert.Equal(t, serviceInfo.Status, uint32(response.Status))
-		assert.Equal(t, v1alpha1.ServiceStatus_STOPPED, response.Status)
+		assert.Equal(t, system.SERVICE_STATUS_STOPPED, response.Status)
```
mauriciopoppe, line 67: Nice, I see that this was generated from protoc but replaced with a const that was already defined in pkg/system/types.go

integrationtests/volume_test.go
```diff
@@ -86,61 +95,334 @@ func runNegativeVolumeStatsRequest(t *testing.T, client *v1beta3client.Client, v
...
-	if client, err = v1beta3client.NewClient(); err != nil {
-		t.Fatalf("Client new error: %v", err)
-	}
-	defer client.Close()
+	client, err := volume.New(volumeapi.New())
+	require.Nil(t, err)
 
 	// Empty volume test
-	runNegativeIsVolumeFormattedRequest(t, client, "")
+	// runNegativeIsVolumeFormattedRequest(t, client, "")
```
mauriciopoppe, line 102: what happened with these tests?

mauriciopoppe (commented):
First pass review:

- Add a more descriptive comment to the PR explaining what were the steps that you took in each commit
- The PR is huge! I'd split it in different PRs, this is so that it's easier to review / traceback changes.

mauriciopoppe:
Let's change the description a little bit more, #217 will become an umbrella item for the overall work i.e. remove fixes #217

ameade:
Looks like this not only enables using the CSI proxy as a library but removes the ability to run it as a gRPC server application? Please let me know if that's not true. Running the CSI Proxy separately from the CSI driver that uses it provides really valuable security isolation. Our current implementation for Windows SMB support (at NetApp) lets us have our Windows pods be unprivileged with no host access, leaving those privileges to the CSI Proxy. This greatly reduces the attack surface.

mauriciopoppe:
@ameade Thanks for the feedback, CSI Proxy does provide format/mount isolation to some directories in the filesystem https://github.com/kubernetes-csi/csi-proxy/blob/v1.x/cmd/csi-proxy/main.go#L49 but with a few drawbacks.

> Running the CSI Proxy separately from the CSI driver that uses it provides really valuable security isolation

- The named pipes are not protected, this means that not only CSI Drivers but any Windows workload can mount them and execute privileged storage operations (imagine a workload reformatting the volume of another workload 😨). This is a current problem as of now.

In addition the current client/server model has some drawbacks, this is one of them:

- It's hard to introduce new changes, to upgrade the CSI Driver with a new version of CSI Proxy we need to install a new binary and restart the Windows service, at the same time if there's a new API the CSI Driver needs to be updated with the new CSI Proxy client, on the CSI Proxy codebase that means supporting lots of autogenerated code like [1](https://github.com/kubernetes-csi/csi-proxy/blob/v1.x/pkg/server/volume/impl/v2alpha1/conversion_generated.go), [2](https://github.com/kubernetes-csi/csi-proxy/blob/v1.x/pkg/server/volume/impl/v2alpha1/server_generated.go) which slows down the development which has gotten to a stable point luckily, there have been attempts to ease the deployment in https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/hostprocess/csi-proxy/README.md but the main problem is still the current client/server model that needs to support different API versions.

By becoming similar to the Linux node component which runs as a privileged container I think we'll have a similar deployment for both OSs. At the same time, enabling HostProcess Pods is also a security concern as it provides **no filesystem isolation**, this is a point I'll cover in the KEP and strategies on how to use it in a cluster.

Let's continue the discussion in the KEP.

mauriciopoppe:
@ameade KEP is up in https://github.com/kubernetes/enhancements/pull/3641, if you have further questions about the motivation, security or any other question please add it to the PR.

PR #228, opened 2022-10-07
Update GitHub action to be branch agnostic

### LINKED ISSUE #222: Update the .github/workflows scripts to work with both v1.x and master
Maintenance of CSI Proxy v1 is going to happen in the v1.x branch, we need to have variations of the windows.yml script that run in v1.x and in master

alexander-ding:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Small update to the windows workflow

Fixes #222 

According to my understanding/reading, an event on GitHub causes a check against the .github/workflows files sitting in the branch associated with it, so we can just set `on: [push, pull_request]`. We then need to merge the changes into the v1.x branches, where the workflows should only run upon pushes/PRs targeting the v1.x branch.  

```release-note
NONE
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,9 +1,5 @@
 name: Windows Tests
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
 jobs:
   integration_tests:
     strategy:


.github/workflows/windows.yml
```diff
@@ -1,9 +1,5 @@
 name: Windows Tests
```
mauriciopoppe, line 1: It looks like master will still run this workflow (and master will soon have the HostProcess code which no longer creates a binary), what's the plan to test HostProcess in master and client/server in v1.x?

alexander-ding: I think the idea is to (a) merge this change in master and then merge it into v1.x (b) slowly add the v2 changes in while keeping v1 around (so the client binary is still there) and (c) delete v1 and the associated v1 GitHub action so that only v2 tests run. During (b) this current test should also run added v2 tests and still pass

mauriciopoppe: sounds like a good plan

PR #230, opened 2022-10-08
Sync v1 to master in preparation for moving master to v2

alexander-ding:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
Syncing v1.x up to master so that we can safely use master to migrate to v2 library model.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #231, opened 2022-10-08
Migrate filesystem API Group to library model

### LINKED ISSUE #226: Refactor the FS API group to become a library

alexander-ding:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Part of #217 (migrating one API Group to v2's library based model). The v1 model will stick around until all API groups are migrate. 

Fixes #226 

Do **not** merge until #230 is merged first.

```release-note
NONE
```
Note that there is no change for now. We don't switch to the new API until the version bump.

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -57,11 +57,7 @@ run_csi_proxy_integration_tests() {
   "& {
     $ErrorActionPreference = \"Stop\";
     Import-Module (Resolve-Path(\"utils.psm1\"));
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestAPIGroups\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestFilesystemAPIGroup\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestDiskAPIGroup\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestVolumeAPIs\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestSmbAPIGroup\";
+    Run-CSIProxyIntegrationTests -test_args \"--test.v\";
   }"
 EOF
 );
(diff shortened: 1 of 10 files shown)


mauriciopoppe (commented):
A lot of these files are showing as new files, I'd move the files in one commit with git mv and then make changes to the imports if needed in another

alexander-ding:
> A lot of these files are showing as new files, I'd move the files in one commit with git mv and then make changes to the imports if needed in another

They're showing up as new files because we're keeping the v1 files around for now. I also think that Git does not track per file history (instead it infers whether a file's been moved by inspecting the diff). I don't think there would be any way around this.

scripts/utils.sh
```diff
@@ -57,11 +57,7 @@ run_csi_proxy_integration_tests() {
   "& {
     $ErrorActionPreference = \"Stop\";
     Import-Module (Resolve-Path(\"utils.psm1\"));
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestAPIGroups\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestFilesystemAPIGroup\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestDiskAPIGroup\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestVolumeAPIs\";
-    Run-CSIProxyIntegrationTests -test_args \"--test.v --test.run TestSmbAPIGroup\";
+    Run-CSIProxyIntegrationTests -test_args \"--test.v\";
```
mauriciopoppe, line 60: I think I did this so that I'd get output for each group independently so that one test failure doesn't skip running the other tests

mauriciopoppe (commented):
I saw this in the logs:

```
=== RUN   TestFilesystem
```

Which means that these tests are running just fine

/lgtm
/approve

scripts/utils.sh
alexander-ding: Ah. This is temporary as we migrate since we have two sets of tests, and it's kind of painful to have to keep adding new lines here to do the v2 tests.

PR #240, opened 2022-10-11
Migrate volume API group to library model

### LINKED ISSUE #235: Refactor the Volume API group to become a library

alexander-ding:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Part of #217 (migrating one API Group to v2's library based model). The v1 model will stick around until all API groups are migrated.

Fixes #235 

```release-note
NONE
```
Deferring release note till all API Groups are migrated.

### DIFF
--- a/integrationtests/volume_test.go
+++ b/integrationtests/volume_test.go
@@ -158,55 +158,6 @@ func TestVolumeAPIs(t *testing.T) {
 	})
 }
 
-// volumeInit initializes a volume, it creates a VHD, initializes it,
-// creates a partition with the max size and formats the volume corresponding to that partition
-func volumeInit(volumeClient volume.Interface, t *testing.T) (*VirtualHardDisk, string, func()) {
-	vhd, vhdCleanup := diskInit(t)
-
-	listRequest := &volume.ListVolumesOnDiskRequest{
-		DiskNumber: vhd.DiskNumber,
-	}
-	listResponse, err := volumeClient.ListVolumesOnDisk(context.TODO(), listRequest)
-	if err != nil {
-		t.Fatalf("List response: %v", err)
-	}
-
-	volumeIDsLen := len(listResponse.VolumeIds)
-	if volumeIDsLen != 1 {
-		t.Fatalf("Number of volumes not equal to 1: %d", volumeIDsLen)
-	}
-	volumeID := listResponse.VolumeIds[0]
-	t.Logf("VolumeId %v", volumeID)
-
-	isVolumeFormattedRequest := &volume.IsVolumeFormattedRequest{
-		VolumeId: volumeID,
-	}
-	isVolumeFormattedResponse, err := volumeClient.IsVolumeFormatted(context.TODO(), isVolumeFormattedRequest)
-	if err != nil {
-		t.Fatalf("Is volume formatted request error: %v", err)
-	}
-	if isVolumeFormattedResponse.Formatted {
-		t.Fatal("Volume formatted. Unexpected !!")
-	}
-
-	formatVolumeRequest := &volume.FormatVolumeRequest{
-		VolumeId: volumeID,
-	}
-	_, err = volumeClient.FormatVolume(context.TODO(), formatVolumeRequest)
-	if err != nil {
-		t.Fatalf("Volume format failed. Error: %v", err)
-	}
-
-	isVolumeFormattedResponse, err = volumeClient.IsVolumeFormatted(context.TODO(), isVolumeFormattedRequest)
-	if err != nil {
-		t.Fatalf("Is volume formatted request error: %v", err)
-	}
-	if !isVolumeFormattedResponse.Formatted {
-		t.Fatal("Volume should be formatted. Unexpected !!")
-	}
-	return vhd, volumeID, vhdCleanup
-}
-
 func getClosestVolumeFromTargetPathTests(diskClient disk.Interface, volumeClient volume.Interface, t *testing.T) {
 	t.Run("DriveLetterVolume", func(t *testing.T) {
 		vhd, _, vhdCleanup := volumeInit(volumeClient, t)
@@ -436,21 +387,6 @@ func mountVolumeTests(diskClient disk.Interface, volumeClient volume.Interface, 
 	}
 }
 
-func volumeTests(t *testing.T) {
-	volumeClient, err := volume.New(volumeapi.New())
-	require.Nil(t, err)
-
-	diskClient, err := disk.New(diskapi.New())
-	require.Nil(t, err)
-
-	t.Run("MountVolume", func(t *testing.T) {
-		mountVolumeTests(diskClient, volumeClient, t)
-	})
-	t.Run("GetClosestVolumeFromTargetPath", func(t *testing.T) {
-		getClosestVolumeFromTargetPathTests(diskClient, volumeClient, t)
-	})
-}
-
 func TestVolume(t *testing.T) {
 	t.Run("NegativeDiskTests", func(t *testing.T) {
 		negativeDiskTests(t)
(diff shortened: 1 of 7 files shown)


integrationtests/volume_test.go
```diff
@@ -177,3 +157,314 @@ func TestVolumeAPIs(t *testing.T) {
 		v2alpha1VolumeTests(t)
 	})
 }
+
+// volumeInit initializes a volume, it creates a VHD, initializes it,
+// creates a partition with the max size and formats the volume corresponding to that partition
+func volumeInit(volumeClient volume.Interface, t *testing.T) (*VirtualHardDisk, string, func()) {
```
mauriciopoppe, line 163: I think this should be part of utils

integrationtests/volume_test.go
```diff
@@ -119,35 +122,12 @@ func negativeVolumeTests(t *testing.T) {
...
-
 func negativeDiskTests(t *testing.T) {
-	var client *v1beta3client.Client
-	var err error
-
-	if client, err = v1beta3client.NewClient(); err != nil {
-		t.Fatalf("Client new error: %v", err)
-	}
-	defer client.Close()
+	_, err := volume.New(volumeapi.New())
```
mauriciopoppe, line 126: below I saw:

```
	volumeClient, err := volume.New(volumeapi.New())
```

which is the same as this test, because we're just moving things around I think it's ok to keep it in this PR but we should remove this test in a later PR when things get moved and stable

pkg/volume/api/api.go
```diff
@@ -0,0 +1,388 @@
+package api
```
mauriciopoppe, line 1: what do you see on git status when you did git mv?

integrationtests/volume_test.go
```diff
@@ -177,3 +157,314 @@ func TestVolumeAPIs(t *testing.T) {
...
+		negativeVolumeTests(t)
+	})
+
+	// TODO: These tests will fail on Github Actions because Hyper-V is disabled
+	// see https://github.com/actions/virtual-environments/pull/2525
+
+	// these tests should be considered frozen from the API point of view
+	t.Run("volumeTests", func(t *testing.T) {
+		skipTestOnCondition(t, isRunningOnGhActions())
+		volumeTests(t)
```
mauriciopoppe, line 468: I think we could include the contents of volumeTests directly here, that way we don't have `TestVolume` and `volumeTests`

mauriciopoppe (commented):
I checked the logs and errors like this one are expected:

```
E1011 22:05:00.850968    4652 volume.go:213] failed GetVolumeStats out [...] outstring Get-Volume : No MSFT_Volume objects found with property 'UniqueId' equal to '-1'.  Verify the value of the property 
```

pkg/volume/api/api.go
alexander-ding: I think it is expected that we're making copies here. We have to keep the old files around. Git doesn't have a separate notion of how the files between commits came to be beyond just intelligently diffing commits and inferring what happened

alexander-ding:
> Is this error ok?
> 
> ```
> === RUN   TestVolume/volumeTests/MountVolume
> E1011 22:05:00.850968    4652 volume.go:213] failed GetVolumeStats out [...] outstring Get-Volume : No MSFT_Volume objects found with property 'UniqueId' equal to '-1'.  Verify the value of the property 
> and retry.
> At line:1 char:51
> + ... ssPreference = 'SilentlyContinue'; (Get-Volume -UniqueId "-1" | Selec ...
> +                                         ~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : ObjectNotFound: (-1:String) [Get-Volume], CimJobException
>     + FullyQualifiedErrorId : CmdletizationQuery_NotFound_UniqueId,Get-Volume
>  
>  err invalid character 'G' looking for beginning of value
> ```
> 
> The last statement makes is sounds like it's an invalid command instead of a command that's trying to get a non-existent volume

This is actually fine (though very confusing). The full error returned is
```
Get-Volume : No MSFT_Volume objects found with property 'UniqueId' equal to 'c'.  Verify the value of the property and retry.   
At line:1 char:51
+ ... essPreference = 'SilentlyContinue'; (Get-Volume -UniqueId 'c' | Selec ...
+                                          ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (c:String) [Get-Volume], CimJobException
    + FullyQualifiedErrorId : CmdletizationQuery_NotFound_UniqueId,Get-Volume
```
which is expected.

The
```err invalid character 'G' looking for beginning of value```
is because Go is trying to read the output message and is getting this error message.

File: pkg/volume/api/api.go [hunk 5]
```diff
@@ -0,0 +1,388 @@
...
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/kubernetes-csi/csi-proxy/pkg/utils"
+	"k8s.io/klog/v2"
+)
+
+// API exposes the internal volume operations available in the server
```
ddebroy, line 16: (Since we are refactoring quite a bit) Instead of calling this `API`, how about something a little more descriptive like `HostAPI` or `HostUtils` for the interface as well as package name?

pkg/volume/volume.go
```diff
@@ -0,0 +1,316 @@
...
+	GetVolumeDiskNumber(context.Context, *VolumeDiskNumberRequest) (*VolumeDiskNumberResponse, error)
+	GetVolumeIDFromMount(context.Context, *VolumeIDFromMountRequest) (*VolumeIDFromMountResponse, error)
+	GetVolumeIDFromTargetPath(context.Context, *GetVolumeIDFromTargetPathRequest) (*GetVolumeIDFromTargetPathResponse, error)
+	GetVolumeStats(context.Context, *GetVolumeStatsRequest) (*GetVolumeStatsResponse, error)
+	IsVolumeFormatted(context.Context, *IsVolumeFormattedRequest) (*IsVolumeFormattedResponse, error)
+	ListVolumesOnDisk(context.Context, *ListVolumesOnDiskRequest) (*ListVolumesOnDiskResponse, error)
+	MountVolume(context.Context, *MountVolumeRequest) (*MountVolumeResponse, error)
+	ResizeVolume(context.Context, *ResizeVolumeRequest) (*ResizeVolumeResponse, error)
+	UnmountVolume(context.Context, *UnmountVolumeRequest) (*UnmountVolumeResponse, error)
+	VolumeStats(context.Context, *VolumeStatsRequest) (*VolumeStatsResponse, error)
```
ddebroy, line 30: Since this API will be for new clients and maintaining perfect compat with prior APIs is not as big of a goal, how about dropping `VolumeStats` (since we already have `GetVolumeStats`)?

pkg/volume/volume.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	return response, nil
+}
+
+func (v *Volume) GetVolumeDiskNumber(context context.Context, request *VolumeDiskNumberRequest) (*VolumeDiskNumberResponse, error) {
+	getDiskNumberFromVolumeIDRequest := &GetDiskNumberFromVolumeIDRequest{
+		VolumeId: request.VolumeId,
+	}
+	getDiskNumberFromVolumeIDResponse, err := v.GetDiskNumberFromVolumeID(context, getDiskNumberFromVolumeIDRequest)
```
ddebroy, line 231: Give this is just a wrapper and around for API compat guarantees of csi-proxy, how about dropping `GetVolumeDiskNumber`?

pkg/volume/volume.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	return response, nil
+}
+
+func (v *Volume) GetVolumeIDFromMount(context context.Context, request *VolumeIDFromMountRequest) (*VolumeIDFromMountResponse, error) {
+	getVolumeIDFromTargetPathRequest := &GetVolumeIDFromTargetPathRequest{
+		TargetPath: request.Mount,
+	}
+	getVolumeIDFromTargetPathResponse, err := v.GetVolumeIDFromTargetPath(context, getVolumeIDFromTargetPathRequest)
```
ddebroy, line 266: Looks like this is compat wrapper as well. Consider dropping `GetVolumeIDFromMount`.

ddebroy (commented):
Overall, the structure looks nice! A few comments about some interface members that I think were around for compat reasons. Those could now be dropped or shortly after in a separate PR.

pkg/volume/volume.go
alexander-ding: Yeah! I just talked to @mauriciopoppe and discovered what these extra methods are. The plan is to have a PR to drop all the compat methods after the initial API group migrations (see #241). I think it's easier to just copy paste the 
current implementation in the first pass.

pkg/volume/api/api.go
alexander-ding: Makes sense. I like HostAPI. How about we do this in a separate PR for all API groups after the initial migration? I like batching similar refactors in one go.

alexander-ding: Tracking this in #242

pkg/volume/volume.go
mauriciopoppe: tracked in https://github.com/kubernetes-csi/csi-proxy/issues/241

pkg/volume/volume.go
mauriciopoppe: tracked in https://github.com/kubernetes-csi/csi-proxy/issues/241

PR #244, opened 2022-10-12
Migrate system API group to library model

### LINKED ISSUE #236: Refactor the System API group to become a library

alexander-ding:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Part of https://github.com/kubernetes-csi/csi-proxy/issues/217 (migrating one API Group to v2's library based model). The v1 model will stick around until all API groups are migrated.

Fixes #236 

```release-note
NONE
```
Deferring release note till all API groups are migrated.

### DIFF
--- a/pkg/system/api/types.go
+++ b/pkg/system/api/types.go
@@ -1,0 +1,12 @@
+package api
+
+type ServiceInfo struct {
+	// Service display name
+	DisplayName string `json:"DisplayName"`
+
+	// Service start type
+	StartType uint32 `json:"StartType"`
+
+	// Service status
+	Status uint32 `json:"Status"`
+}
--- a/integrationtests/system_test.go
+++ b/integrationtests/system_test.go
@@ -8,8 +8,12 @@ import (
 	"strings"
 	"testing"
 
+	system "github.com/kubernetes-csi/csi-proxy/pkg/system"
+	systemapi "github.com/kubernetes-csi/csi-proxy/pkg/system/api"
+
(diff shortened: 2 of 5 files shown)


pkg/system/api/types.go
```diff
@@ -0,0 +1,12 @@
+package api
```
mauriciopoppe, line 1: where is this file coming from?

alexander-ding: It's defined here inline in the [old OS file](https://github.com/alexander-ding/csi-proxy/blob/015225affc1e207eb931fec055653a407c8c3688/pkg/os/system/api.go#L17). Since this is user facing, it should really get its own definition. I just copied over the struct from API so that the internal impl and the PKG have their own versions.

mauriciopoppe: is it possible to move this to pkg/system/types.go?

alexander-ding: My understanding is that it's bad practice to return non-exported types in exported functions/methods. For example, if a client wants to implement the API interface in pkg/system/api.go (for whatever reason; maybe mocking), then they would want to be able to specify the type. This wouldn't be possible if the type isn't exported.

mauriciopoppe: sure, I was trying to minimize what's visible to the client but as you mention there are practical reasons to have it like this

PR #247, opened 2022-10-13
Migrate iscsi API group to library model

### LINKED ISSUE #238: Refactor the iSCSI API group to become a library

alexander-ding:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #238 

Removed the tests for the v1 API since it doesn't follow the versioned API like other API groups, which makes having two sets of tests a lot more work. We're removing v1 soon anyways so this is probably fine.

```release-note
NONE
```
Deferring release note to the version bump PR.

/cc @mauriciopoppe


pkg/disk/disk.go
```diff
@@ -17,13 +17,29 @@ type Interface interface {
 	DiskStats(context.Context, *DiskStatsRequest) (*DiskStatsResponse, error)
 	GetAttachState(context.Context, *GetAttachStateRequest) (*GetAttachStateResponse, error)
 	GetDiskNumberByName(context.Context, *GetDiskNumberByNameRequest) (*GetDiskNumberByNameResponse, error)
+
+	// GetDiskState gets the offline/online state of a disk.
```
mauriciopoppe, line 21: I think this could be part of a different PR, I'd limit this change to iSCSI changes only

vendor/github.com/kubernetes-csi/csi-proxy/client/groups/iscsi/v1alpha2/client_generated.go
```diff
@@ -1,98 +0,0 @@
-// Code generated by csi-proxy-api-gen. DO NOT EDIT.
```
mauriciopoppe, line 1: Files are getting removed here while in previous PRs it was only about adding changes, is this intended or should this happen later?

pkg/disk/disk.go
alexander-ding: Oh oops I meant to leave these changes for later. Accidentally pushed them.

vendor/github.com/kubernetes-csi/csi-proxy/client/groups/iscsi/v1alpha2/client_generated.go
alexander-ding: Yes that's intended. See the note
> Removed the tests for the v1 API since it doesn't follow the versioned API like other API groups, which makes having two sets of tests a lot more work. We're removing v1 soon anyways so this is probably fine.
This means that none of the pkg code uses the v1 client anymore, so it causes the vendor dependency to be deleted.

mauriciopoppe: ack

mauriciopoppe:
I think you could try squashing your changes and then we can merge it

alexander-ding:
> I think you could try squashing your changes and then we can merge it

Done! Thanks for the review :)

mauriciopoppe (commented):
/lgtm
/approve

PR #257, opened 2022-10-17
Remove compatibility methods for old API versions

### LINKED ISSUE #241: Remove v1beta and v1alpha compatibility methods in the server side for the API groups that are GA

alexander-ding:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

As we migrate to v2, old compatibility methods no longer need to be maintained by the server.

Fixes #241 

```release-note
NONE
```
Deferring release note till version bump.

/cc @mauriciopoppe


mauriciopoppe:
I think we should also add in the migration guide some notes about moving from a v1alphaX, v1betaX version to the library functions, although I think most of the CSI Drivers that support Windows should already be using v1betaX or v1

mauriciopoppe (commented):
Thanks for removing unused code!

/lgtm
/approve

PR #262, opened 2022-10-18
Delete Makefile checks from GitHub actions

### LINKED ISSUE #259: GitHub actions running go fmt check on a Windows machine complains about file format endings
**What happened**:
As discussed in [this PR](https://github.com/kubernetes-csi/csi-proxy/pull/258#issuecomment-1281744970), the automatic GitHub action for Windows unit tests is running on a Windows machine, where `gofmt` expects \n\r line endings in files. However, the files checked in in this repository use Linux style \n line endings, which causes `gofmt` to complain. This doesn't show up in local Linux development machines and have been silent as GitHub actions only care about the return code of the final command in a multi-line run command. The `gofmt` check is only the first line, so the error is logged but doesn't fail the entire action.

The `gofmt` check, as well as other Makefile checks, should happen on a Linux machine. The actual unit tests can happen in its own separate task.

**How to reproduce it**:
Any pull request onto master should trigger the workflow. Check the log for the unit tests. 

**Environment**:
Not relevant.

alexander-ding:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
/kind failing-test
> /kind feature
> /kind flake

GitHub action test is overlapping with Prow. 

Fixes #259 

Reopening to trigger the GH action job again.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -33,5 +33,4 @@ jobs:
         uses: actions/checkout@v2
       - name: Run Windows Unit Tests
         run: |
-          make all
           go test -v -race ./pkg/...


.github/workflows/makefile.yml
```diff
@@ -0,0 +1,21 @@
+name: Makefile Checks
```
mauriciopoppe, line 1: Please rename this file to code format checks, also add a comment at the top explaining why we need a linux job in addition to the Windows job

alexander-ding: Done! Force-pushed to squash.

.github/workflows/code_format_checks.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Run Code Format Checks
+        run: |
+          make test-fmt &&
```
mauriciopoppe, line 19: there was the subtree check too right? I think this test that release-tools wasn't modified and that I think is important since this folder comes from a different repo.

could we run `make all` as it is in Linux instead?

alexander-ding: `test-subtree` was failing in my previous attempt to refactor the GitHub actions: PR #261. In fact I think this issue would have been happening all along, but the error was not showing because `make all` fails on `test-fmt` and never got to execute `test-subtree`. 

The reason behind this is that GitHub Action's checkout@v2 grabs the latest commit as a graft, which causes the `git log -- release-tools` command run by verify-subtree to think that the last commit added the entire repository (or something like that).

I did some digging, and it seems like no other repo in CSI Proxy runs verify-subtree in a GitHub action. I think we can just remove it and run `make test-fmt test-vendor test-shellcheck`.

.github/workflows/code_format_checks.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
mauriciopoppe, line 20: lgtm, I was reading the logs and there's this job which is already doing the same I think `pull-kubernetes-csi-csi-proxy`, I didn't check the full log though

alexander-ding: I think you're right. Should we just delete this then?

mauriciopoppe: yes please, removing `make all` from the windows job is great though

PR #274, opened 2022-10-24
test: migrate tests to HostProcess pods

### LINKED ISSUE #227: Update the integration tests to run in a HostProcess pod

alexander-ding:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

As we migrate to a library model, the tests need to more closely mimic the runtime environment of the code, which is in HostProcess containers. Thus, this PR updates the tests to spin up a pod running HostProcess containers to execute the tests, instead of directly running the test binary on the Windows node.

Fixes #227

```release-note
NONE
```

/cc @mauriciopoppe

### DIFF
--- a/scripts/integration-test.yaml
+++ b/scripts/integration-test.yaml
@@ -1,0 +1,27 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: integration-test
+  labels:
+    app: integration-test
+spec:
+  nodeName: windows_node # to be replaced by sed
+  securityContext:
+    windowsOptions:
+      hostProcess: true
+      runAsUserName: "NT AUTHORITY\\SYSTEM"
+  hostNetwork: true
+  containers:
+  - name: agnhost
+    image: mcr.microsoft.com/windows/nanoserver:ltsc2019
+    command: ['powershell', '-command', 'Start-Sleep 100000']
+    volumeMounts:
+      - mountPath: /integration-test
+        name: integration-test-volume
+  nodeSelector:
+    kubernetes.io/os: windows
+  volumes:
+    - name: integration-test-volume
+      hostPath:
+        path: /
+        type: DirectoryOrCreate
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -16,7 +16,7 @@ if ! [ -z "${GCP_ZONE:-}" ]; then
 fi
 
 sync_file_to_vm() {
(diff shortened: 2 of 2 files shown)


scripts/integration-test.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: integration-test
+  labels:
+    app: integration-test
+spec:
+  nodeName: windows_node # to be replaced by sed
+  securityContext:
+    windowsOptions:
+      hostProcess: true
+      runAsUserName: "NT AUTHORITY\\SYSTEM"
```
mauriciopoppe, line 12: let's experiment here, could you try the user with the lowest privilege first?

scripts/utils.sh
```diff
@@ -33,12 +37,13 @@ sync_csi_proxy_integration_tests() {
 run_csi_proxy_integration_tests() {
   echo "Run integration tests"
   local ps1=$(cat << 'EOF'
-  "& {
-    $ErrorActionPreference = \"Stop\";
-    .\integrationtests.test.exe --test.v
-  }"
+    $ErrorActionPreference = "Stop";
```
mauriciopoppe, line 40: did it fail when it was in its own scope e.g. when inside `&{ }`

alexander-ding: The error code doesn't propagate correctly when in a subshell. We want the overall command to fail if the test fails.

scripts/integration-test.yaml
alexander-ding: In my previous experiments, anything less than SYSTEM hasn't worked. Let me try again locally and get back to you though.

alexander-ding: Yeah, even setting the username to `"NT AUTHORITY\\Local service"` causes commands like 
```
C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe /c & { $global:ProgressPreference = 'SilentlyContinue'; (Get-VHD -Path C:\\var\\lib\\kubelet\\plugins\\testplugin-693308.csi.io\\disk-693308.vhdx).DiskNumber }
```
to fail with permission issues.

alexander-ding: And Network service fails too

mauriciopoppe (commented):
/lgtm
/approve

PR #275, opened 2022-10-25
Bump version to v2

alexander-ding:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

```release-note
action required: releases CSI Proxy v2

As Windows HostProcess pods reach beta in Kubernetes 1.23, we no longer need the clients/server model of CSI Proxy to run privileged storage operations. Driver deployments can run as HostProcess containers to directly issue the privileged commands. As such, CSI Proxy is becoming a format/mount utility library similar to [kubernetes/mount-utils](https://github.com/kubernetes/mount-utils) that CSI Drivers can include to get convenience methods to execute privileged storage operations.

We have also taken advantage of the major version change to reorganize the package structure and improve some API names. Details on this version bump and a migration guide will be available in the KEP and corresponding blog post upon official release. In the mean time, we will provide usage instructions for CSI Proxy v2 in [/docs/USAGE.md](https://github.com/kubernetes-csi/csi-proxy/blob/library-development/docs/USAGE.md) on this branch for driver developers to experiment with the migration.

Below is a list of breaking changes:

1. The host binary is no longer generated. CSI Proxy becomes a regular Go library.
2. Individual API group versioning is removed. CSI Proxy maintains a single notion of version via Go module version.
3. The import path needs to include `/v2` after the repository path due to the version bump (e.g., `github.com/kubernetes-csi-proxy/v2/pkg/...`).
4. Each API group is now available under `/pkg/<API group>` as an interface defining the API and a struct implementing the interface. A `New` method is exposed under that path to initialize an instance of the struct, and it always takes in a corresponding `HostAPI` interface (previously called `API`), which is responsible for communicating with the OS, available under `/pkg/<API group>/hostapi`. The `/pkg/<API group>/hostapi` package follows a similar pattern, exposing an interface, an implementing struct, and a `New` method to initialize an instance of the struct. Previously, not every API group followed this pattern.
5. Every API group method `DoFoo` now takes in a request struct `&DoFooRequest{...}` and returns a response struct `&DoFooResponse{...}`. Previously, not every API group followed this pattern.
6. Compatibility methods to accommodate legacy API group versions are removed. Only the latest version of each API group is supported. This will continue to be the pattern in future releases of v2.
7. Go variable/struct names containing acronyms are fully capitalized, following Go style guides. Across the board, `Id`'s are renamed to `ID`s; `Smb` is renamed to `SMB`; `Iqn` is renamed to `IQN`; and `iscsi` is renamed to `iSCSI`.
8. Drivers importing CSI Proxy v2 must be deployed as HostProcess containers. See the above usage guide for migration details and caveats.
```

/cc @mauriciopoppe

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/kubernetes-csi/csi-proxy
+module github.com/kubernetes-csi/csi-proxy/v2
 
 go 1.16
 
--- a/integrationtests/disk_test.go
+++ b/integrationtests/disk_test.go
@@ -9,8 +9,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kubernetes-csi/csi-proxy/pkg/disk"
-	diskapi "github.com/kubernetes-csi/csi-proxy/pkg/disk/hostapi"
+	"github.com/kubernetes-csi/csi-proxy/v2/pkg/disk"
+	diskapi "github.com/kubernetes-csi/csi-proxy/v2/pkg/disk/hostapi"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
--- a/integrationtests/filesystem_test.go
+++ b/integrationtests/filesystem_test.go
@@ -14,8 +14,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/kubernetes-csi/csi-proxy/pkg/filesystem"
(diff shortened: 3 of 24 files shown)


mauriciopoppe:
Thanks for bumping CSI Proxy to v2, I'll add a note in the KEP about this too.

/lgtm
/approve

PR #292, opened 2023-07-25
Bump to go 1.20

mauriciopoppe:
**What type of PR is this?**
>
/kind cleanup

Bump to go 1.20, set dependency management tooling. 

```release-note
NONE
```

/assign @msau42 
/cc @marosset


.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Install go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ^1.20
+
+      - name: Run Trivy vulnerability scanner in repo mode
```
msau42, line 20: We have been thinking about using Github's built-in security scanning for csi repos instead of trivy. Do you see any advantages for using trivy?

mauriciopoppe: Not really, I thought we didn't have any sort of visibility on vulns so I copied https://github.com/kubernetes-csi/external-snapshotter/blob/master/.github/workflows/trivy.yaml to this repo but if we already have something built in then yes I think we can use that instead.

mauriciopoppe: I removed the file.

PR #315, opened 2023-09-27
Add require-privacy option for New-SmbGlobalMapping (default: true)

vitaliy-leschenko:
**What type of PR is this?**
/kind feature

Currently New-SmbGlobalMapping is called always with option `-RequirePrivacy $true`. It might be a reason for the issue when SMB server doesn't support encryption. This PR allows to change this behavior by adding new flag (-require-privacy) to csi-proxy.exe. It is true by default for backward compatibility. If we want to allow call New-SmbGlobalMapping without -RequirePrivacy $true just add `-require-privacy=false` to csi-proxy arguments.

```release-note
The flag (-require-privacy) has been added. If true, New-SmbGlobalMapping will be called with -RequirePrivacy $true
```

### DIFF
--- a/cmd/csi-proxy/main.go
+++ b/cmd/csi-proxy/main.go
@@ -34,10 +34,11 @@ func (i *workingDirFlags) Set(value string) error {
 }
 
 var (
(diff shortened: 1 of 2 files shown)


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #348, opened 2024-06-26
Update the github action workflow to run the integration tests

mauriciopoppe:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
/kind bug

**What this PR does / why we need it**:
A presubmit is failing because of a change in the github actions default Windows VM https://github.blog/changelog/2022-01-11-github-actions-jobs-running-on-windows-latest-are-now-running-on-windows-server-2022/, I used the same code in `scripts/run-integration.sh` rewritten in powershell.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #353, opened 2024-08-20
Remove --kube-root deprecated kind argument

stmcginnis:
**What type of PR is this?**

/kind cleanup

The `--kube-root` flag has been deprecated in kind for some time and will be removed. The current equivalent functionality is to pass the path to the kubernetes source as the first non-flag argument to `build node-image`.

This updates the prow script to use the newer preferred method in preparation for kind dropping the flag.

Related: https://github.com/kubernetes-sigs/kind/issues/3717

```release-note
NONE
```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -610,7 +610,7 @@ start_cluster () {
             go_version="$(go_version_for_kubernetes "${CSI_PROW_WORK}/src/kubernetes" "$version")" || die "cannot proceed without knowing Go version for Kubernetes"
             # Changing into the Kubernetes source code directory is a workaround for https://github.com/kubernetes-sigs/kind/issues/1910
             # shellcheck disable=SC2046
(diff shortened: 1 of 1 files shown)


mauriciopoppe:
Thanks for the PR, the integration test is currently failing and it'd prevent merging this PR, I haven't had the time to continue debugging https://github.com/kubernetes-csi/csi-proxy/pull/348 but once that one is merged.

`release-tools` comes from https://github.com/kubernetes-csi/csi-release-tools so this change should go that project to then adopt it in this repo.

stmcginnis:
Thanks @mauriciopoppe - I didn't realize the source at first, but I have submitted https://github.com/kubernetes-csi/csi-release-tools/pull/259 to do the update there. It looks like there are also unrelated test failures, but hopefully that can be resolved soon.

TerryHowe (approved):
/lgtm

stmcginnis:
Change has now merged to csi-release-tools. I'm not sure what the process is to get that change into the other repos that use it. Is there a sync process that needs to happen, or is it a manual update process where this PR would still be needed.

If someone from the SIG could point to any next steps that need to happen, I'm happy to follow through to get this change propagated to everywhere it needs to be.

mauriciopoppe:
It'd be through a change like https://github.com/kubernetes-csi/csi-proxy/pull/362. I think we should superseed this PR by that one.

PR #336, opened 2024-04-16
Failing closed after maximum retry is achieved to avoid inf recursion

knabben:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
/kind bug

**What this PR does / why we need it**:
Adding a maximum retry for the CSI to find the volume target from a mountpath.

**Which issue(s) this PR fixes**:

Fixes #193 

**Special notes for your reviewer**:
Tracking the recursion call trace here:
https://github.com/kubernetes-csi/csi-proxy/issues/193#issuecomment-2059665463

**Does this PR introduce a user-facing change?**:

```release-note
Maximum retry (3) when for the getTarget method iterate and find the correct volume of the mount path. 
```

PR #356, opened 2024-09-23
bypass smbtest to unblock integration tests

Zhupku:
**What type of PR is this?**
>
/kind bug

The SMB test keeps failing. It failed at using "ConvertTo-SecureString", this is because the invocation of pwsh -> cmd -> powershell confuse the $env:PSModulePath. After reset its value, this bug is fixed. However, the test still fails at 
image

Investigating on how to fix this bug. Bypass the smb test temporarily to unblock integration tests.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -4,7 +4,7 @@ jobs:
   integration_tests:
     strategy:
       matrix:
-        go: ['1.20']
+        go: ['1.22']
         platform: [windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
@@ -35,7 +35,7 @@ jobs:
   unit_tests:
     strategy:
       matrix:
-        go: ['1.20']
+        go: ['1.22']
         platform: [windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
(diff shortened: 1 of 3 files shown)


mauriciopoppe:
If the test is disabled how do we make sure that we're not adding a regression? I'm aware this is blocking things like https://github.com/kubernetes-csi/csi-proxy/pull/336 but I think we should attempt to fix the setup not just ignore the tests.

andyzhangx (approved):
/ok-to-test

andyzhangx:
@mauriciopoppe there is sth. wrong with the integration tests:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-csi_csi-proxy/356/pull-kubernetes-csi-csi-proxy-integration/1845716316082671616

shall we merge this PR first?

mauriciopoppe:
This was a good hint: `this is because the invocation of pwsh -> cmd -> powershell confuse the $env:PSModulePath`. I wrote `shell: powershell` to avoid using pwsh in https://github.com/kubernetes-csi/csi-proxy/pull/348 and that seemed to help. Thanks.

mauriciopoppe:
It'll be fixed in https://github.com/kubernetes-csi/csi-proxy/pull/348.

PR #360, opened 2024-10-22
Use WMI to implement Volume API to reduce PowerShell overhead

### LINKED ISSUE #193: High cpu usage of powershell processes triggered by csi-proxy
**What happened**: 
 
**How to reproduce it**:

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

laozc:
/kind feature

This PR leverages of WMI interfaces to replace the PowerShell functions for Windows related storage operations,
which improve the overall performance of csi-proxy.

Ran some test with two Windows nodes setup, one with the CSI-proxy with PowerShell, while another one with the CSI-proxy using WMI.
The test creates two pods side by side and delete the resources from the cluster once they're running.

* Kubernetes: version 1.31.4
* CSI: vSphere CSI
* CNI: Antrea
* Windows Node Spec: 4 vCPU, 16GB memory

Here is the performance data with collected from the gRPC server metrics within 10 minutes time frame.

| Operation | Before (sec) | After (sec) | Before (Min) | After (Max) | Improvement |
|-----------|--------------|-------------|--------------|-------------|-------------|
| FormatVolume (1G disk) | 5.85-8.85 | 2.90s | 5.85 | 2.90 | -101% |
| GetDiskNumberFromVolumeID | 2.90 | 0.29-0.58 | 2.90 | 0.58 | -400% |
| GetVolumeIDFromTargetPath | 0.58-0.96 | 0.009 | 0.58 | 0.009 | -6344% |
| GetVolumeStats | 2.90 | 0.29 | 2.90 | 0.29 | -900% |
| IsVolumeFormatted | 2.90 | 0.0955 | 2.90 | 0.0955 | -2937% |
| ListDiskIDs | 2.9-5.85 | 0.585 | 2.90 | 0.585 | -396% |
| ListVolumeOnDisk | 2.9 | 0.29-0.585 | 2.90 | 0.585 | -396% |
| MountVolume | 2.9-5.85 | 0.00955-0.0955 | 2.90 | 0.09 | -3122% |
| PartitionDisk (1G disk) | 19.5 | 2.9 | 19.5 | 2.9 | -572% |
| SetDiskState | 2.9-5.70 | 0.0955-0.280 | 2.90 | 0.280 | -936% |
| UnmountVolume | 5.85-8.85 | 0.29-0.585 | 5.85 | 0.585 | -900% |

It's obviously shown that PowerShell takes around 2.9s to load in the environment while memory consumption could reach up to 200MB, which becomes the bottle neck for volume mount on Windows.
By reducing the PowerShell overhead, we were able to cut down the time cost significantly to method calls for all operations.

Ref #193

```release-note
NONE
```


andyzhangx:
thanks for the PR. @laozc have you tested the WMI based operation manually? does it work?

laozc:
Yes. I verified it in my local environment and they work the same as PowerShell commands.
I'm still working on finalizing the PR with more testing and integrate it into CSI driver to get some result.

laozc:
Ran some test with two Windows nodes setup, one with the CSI-proxy with PowerShell, while another one with the CSI-proxy using WMI.
The test creates two pods side by side and delete the resources from the cluster once they're running.

* Kubernetes: version 1.31.4
* CSI: vSphere CSI
* CNI: Antrea
* Windows Node Spec: 4 vCPU, 16GB memory

Here is the performance data with collected from the gRPC server metrics within 10 minutes time frame.

| Operation | Before (sec) | After (sec) | Before (Min) | After (Max) | Improvement |
|-----------|--------------|-------------|--------------|-------------|-------------|
| FormatVolume (1G disk) | 5.85-8.85 | 2.90s | 5.85 | 2.90 | -101% |
| GetDiskNumberFromVolumeID | 2.90 | 0.29-0.58 | 2.90 | 0.58 | -400% |
| GetVolumeIDFromTargetPath | 0.58-0.96 | 0.009 | 0.58 | 0.009 | -6344% |
| GetVolumeStats | 2.90 | 0.29 | 2.90 | 0.29 | -900% |
| IsVolumeFormatted | 2.90 | 0.0955 | 2.90 | 0.0955 | -2937% |
| ListDiskIDs | 2.9-5.85 | 0.585 | 2.90 | 0.585 | -396% |
| ListVolumeOnDisk | 2.9 | 0.29-0.585 | 2.90 | 0.585 | -396% |
| MountVolume | 2.9-5.85 | 0.00955-0.0955 | 2.90 | 0.09 | -3122% |
| PartitionDisk (1G disk) | 19.5 | 2.9 | 19.5 | 2.9 | -572% |
| SetDiskState | 2.9-5.70 | 0.0955-0.280 | 2.90 | 0.280 | -936% |
| UnmountVolume | 5.85-8.85 | 0.29-0.585 | 5.85 | 0.585 | -900% |

It's obviously shown that PowerShell takes around 2.9s to load in the environment while memory consumption could reach up to 200MB.
By reducing the PowerShell overhead, we were able to cut down the time cost significantly to method calls for all operations.

mauriciopoppe (commented):
Awesome work! Is your plan to see https://github.com/kubernetes-csi/csi-proxy/pull/371 merged first?

mauriciopoppe:
One more thing, this PR is against the server version of CSI Proxy, we did lots of development against the library version in https://github.com/kubernetes-csi/csi-proxy/tree/library-development which doesn't use a client/server model anymore, I think it'd be great to see this work in that other branch once there's signal that the PR doesn't cause any regressions.

I see that you tested this against the vSphere CSI Driver, do you know if there are plans for that CSI Driver to use the library version of CSI Proxy?

laozc:
For the current 1.x release when gRPC is used, I believe we could merge #371 first in case monitoring the gRPC operation performance metrics is preferred, especially we need some concrete evidence for the case like this PR.

I believe vSphere CSI driver is working on moving to CSI proxy library release as well.
I will be working on the same PR for the library version. Once the PR for the library release is ready, we could safely process with both PRs in two branches without causing any regression.

mauriciopoppe:
Now that https://github.com/kubernetes-csi/csi-proxy/pull/371 let's continue this work! My only concern is that Powershell is way easier to debug than doing syscalls.

Would it be possible to split the PR into multiple PRs grouping by changes per API please? I think it'd be easier to read and review.

laozc:
No problem. Which branch do you think we should target first?

mauriciopoppe:
Let's try on master first, after a new CSI Proxy release it'd be a matter of updating the CSI Proxy binary deployment only, no changes in the CSI Driver.

For the library version, it requires some changes in the CSI Driver.

laozc:
The commit which contains all API changes in the same PR was https://github.com/kubernetes-csi/csi-proxy/commit/6986e5989e4991cd0369396314f13c755d2f5309

laozc:
I believe we could safely remove the constraint to use Go 1.23 with this change.
```
mountedFolder, err := isMountedFolder(candidatePath)
```
This method uses Windows API calls to determine whether the path is a mounted path instead of relying on the Go API behavior.

Makefile
```diff
@@ -15,8 +15,6 @@ CSI_PROXY_API_GEN = $(BUILD_DIR)/csi-proxy-api-gen
 # overrides the $(CMDS:%=build-%) rule in release-tools/build.make because this project is not compatible with go >1.23
 # TODO: remove this override as part of https://github.com/kubernetes-csi/csi-proxy/issues/361
 build-csi-proxy: check-go-version-go
-	# Checks that the go version is 1.22 or lower
```
mauriciopoppe, line 18: I added this check so that during the build of CSI Proxy if the environment uses >go1.22 then it'd break.

At the same time go.mod doesn't seem to have any changes, the min go version is still go1.22.

Is this PR enough to say that all the privileged storage operations are no longer going to fail with go1.23+? I think that moving to go1.23+ could be a cleanup item after all the refactors.

pkg/os/volume/api.go
```diff
@@ -1,15 +1,18 @@
 package volume
 
 import (
-	"encoding/json"
+	errorspkg "errors"
```
mauriciopoppe, line 4: I think this should be "errors" only without a named import

pkg/os/volume/api.go
```diff
@@ -1,15 +1,18 @@
...
 	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
 
-	"github.com/kubernetes-csi/csi-proxy/pkg/utils"
+	"github.com/go-ole/go-ole"
+	"github.com/kubernetes-csi/csi-proxy/pkg/cim"
+	"github.com/microsoft/wmi/pkg/errors"
```
mauriciopoppe, line 14: the named import might come here.

pkg/os/volume/api.go
```diff
@@ -54,56 +57,67 @@ var (
...
-	cmdEnv := fmt.Sprintf("volumeID=%s", volumeID)
-	out, err := utils.RunPowershellCmd(cmd, cmdEnv)
-
-	if err != nil || len(out) == 0 {
-		return -1, fmt.Errorf("error getting size of the partition from mount. cmd %s, output: %s, error: %v", cmd, string(out), err)
+// ListVolumesOnDisk - returns back list of volumes(volumeIDs) in a disk and a partition.
+func (VolumeAPI) ListVolumesOnDisk(diskNumber uint32, partitionNumber uint32) (volumeIDs []string, err error) {
+	partitions, err := cim.ListPartitionsOnDisk(diskNumber, partitionNumber, []string{"ObjectId"})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to list partition on disk %d", diskNumber)
```
mauriciopoppe, line 75: I believe this would be on top of "errors" and not "github.com/microsoft/wmi/pkg/errors" after the refactor.

pkg/cim/volume.go
```diff
@@ -0,0 +1,241 @@
+package cim
```
mauriciopoppe, line 1: I believe only methods from this file are used in pkg/os/volume/api.go. Would it be possible to add the other files in pkg/cim in a different PR?

mauriciopoppe (commented):
I left some comments.

Makefile
laozc: Sure. The second commit was discarded.

pkg/cim/volume.go
laozc: The functions in `cim` package are grouped by WMI classes.
disk.go and volume.go with required functions are kept for this Volume API PR.
I will do a rebase on the other PRs.

mauriciopoppe:
Hey @andyzhangx, would it be possible for you to review the PR please? Maybe you have experience with the `github.com/microsoft/wmi` package and how to use it. From my end, I can only check that the integration tests are still passing.

pkg/cim/volume.go
```diff
@@ -0,0 +1,297 @@
...
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/microsoft/wmi/pkg/base/query"
+	"github.com/microsoft/wmi/pkg/errors"
+	cim "github.com/microsoft/wmi/pkg/wmiinstance"
+	"github.com/microsoft/wmi/server2019/root/microsoft/windows/storage"
```
andyzhangx, line 10: this relies on specific server version, does it work on server 2022 and 2025?

andyzhangx (commented):
/hold
I will try this PR on Azure CSI driver first to make sure it works on common Windows server versions.

andyzhangx:
@laozc I tries to copy the cim folder into my Azure Disk CSI driver repo (for test): https://github.com/kubernetes-sigs/azuredisk-csi-driver/compare/master...andyzhangx:azuredisk-csi-driver:cim?expand=1, but I got following error, have you hit similar error in your CSI driver?

https://github.com/andyzhangx/azuredisk-csi-driver/actions/runs/14463170910/job/40559504163

```
# sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim
FAIL	sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim [setup failed]
package sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim
	imports github.com/microsoft/wmi/pkg/wmiinstance
	imports golang.org/x/sys/windows: build constraints exclude all Go files in /home/runner/work/azuredisk-csi-driver/azuredisk-csi-driver/vendor/golang.org/x/sys/windows
# sigs.k[8](https://github.com/andyzhangx/azuredisk-csi-driver/actions/runs/14463170910/job/40559504163#step:4:9)s.io/azuredisk-csi-driver/pkg/os/cim
package sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim
	imports github.com/microsoft/wmi/server201[9](https://github.com/andyzhangx/azuredisk-csi-driver/actions/runs/14463170910/job/40559504163#step:4:10)/root/microsoft/windows/storage
	imports github.com/microsoft/wmi/pkg/base/instance
	imports github.com/microsoft/wmi/pkg/base/session: build constraints exclude all Go files in /home/runner/work/azuredisk-csi-driver/azuredisk-csi-driver/vendor/github.com/microsoft/wmi/pkg/base/session
```

pkg/cim/volume.go
laozc: I always tested on Windows Server 22H2 with this API.
I don't think it would break as long as the new feature introduced in newer API is not used.

laozc:
> @laozc I tries to copy the cim folder into my Azure Disk CSI driver repo (for test): https://github.com/kubernetes-sigs/azuredisk-csi-driver/compare/master...andyzhangx:azuredisk-csi-driver:cim?expand=1, but I got following error, have you hit similar error in your CSI driver?
> 
> https://github.com/andyzhangx/azuredisk-csi-driver/actions/runs/14463170910/job/40559504163
> 
> ```
> # sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim
> FAIL	sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim [setup failed]
> package sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim
> 	imports github.com/microsoft/wmi/pkg/wmiinstance
> 	imports golang.org/x/sys/windows: build constraints exclude all Go files in /home/runner/work/azuredisk-csi-driver/azuredisk-csi-driver/vendor/golang.org/x/sys/windows
> # sigs.k[8](https://github.com/andyzhangx/azuredisk-csi-driver/actions/runs/14463170910/job/40559504163#step:4:9)s.io/azuredisk-csi-driver/pkg/os/cim
> package sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim
> 	imports github.com/microsoft/wmi/server201[9](https://github.com/andyzhangx/azuredisk-csi-driver/actions/runs/14463170910/job/40559504163#step:4:10)/root/microsoft/windows/storage
> 	imports github.com/microsoft/wmi/pkg/base/instance
> 	imports github.com/microsoft/wmi/pkg/base/session: build constraints exclude all Go files in /home/runner/work/azuredisk-csi-driver/azuredisk-csi-driver/vendor/github.com/microsoft/wmi/pkg/base/session
> ```

Hi, these Go files in `github.com/microsoft/wmi/pkg/base/instance` and `golang.org/x/sys/windows` are set with a build flat to only work under `GOOS=windows` so they should be excluded from UT.
I don't think the code to call WMI pkg/os/cim could be tested in UT stage as they're wrapping WMI.
The E2E cases should cover the functionality.

You may either
`go -covermode=atomic -coverprofile=profile.cov $(go list ./... | grep -v sigs.k8s.io/azuredisk-csi-driver/pkg/os/cim)` to exclude these package
or either add
```
//go:build windows
// +build windows
```
to each file in this package so the test coverage is skipped for these files on Linux.

laozc:
I could add build constraints to these codes if the downstream projects need.
```
//go:build windows
// +build windows
```

andyzhangx:
> //go:build windows
> // +build windows

yes, that's necessary, and pls also add header, you could refer to 
https://github.com/kubernetes-sigs/azuredisk-csi-driver/commit/d442711d9b5f9b9bd77c5337d061255e18210cc5

pkg/os/volume/api.go
```diff
@@ -149,92 +177,117 @@ func (VolumeAPI) UnmountVolume(volumeID, path string) error {
...
-	cmdEnv = append(cmdEnv, fmt.Sprintf("volumeID=%s", volumeID))
-	cmdEnv = append(cmdEnv, fmt.Sprintf("mountpath=%s", path))
-	out, err := utils.RunPowershellCmd(cmd, cmdEnv...)
-
+	mountPoint := path
+	if !strings.HasSuffix(mountPoint, "\\") {
+		mountPoint += "\\"
+	}
+	utf16MountPath, _ := windows.UTF16PtrFromString(mountPoint)
+	err := windows.DeleteVolumeMountPoint(utf16MountPath)
```
andyzhangx, line 185: is this identical to `Remove-PartitionAccessPath` command?

andyzhangx (commented):
/hold
I am trying the new cim calls

pkg/os/volume/api.go
```diff
@@ -149,92 +177,117 @@ func (VolumeAPI) UnmountVolume(volumeID, path string) error {
...
 		return fmt.Errorf("error getting the current size of volume (%s) with error (%v)", volumeID, err)
 	}
 
+	currentSize, err := strconv.ParseInt(currentSizeVal.(string), 10, 64)
+	if err != nil {
+		return fmt.Errorf("error parsing the current size of volume (%s) with error (%v)", volumeID, err)
+	}
+
 	// only resize if finalSize - currentSize is greater than 100MB
 	if finalSize-currentSize < 100*1024*1024 {
```
andyzhangx, line 230:
```
	if finalSize-currentSize < 100*1024*1024 {
		klog.V(2).Infof("minimum resize difference(1GB) not met, skipping resize. volumeID=%s currentSize=%d finalSize=%d", volumeID, currentSize, finalSize)
		return nil
	}
```

andyzhangx: just add `return nil` in this condition, that's one regression I have been found in our disk csi driver e2e test failures.

andyzhangx:
@laozc do you have the link about how this WMI improves the performance? pls paste details in this PR description, thanks.

pkg/os/volume/api.go
laozc: I believe so.
The `Add-PartitionAccessPath` should be identical to `SetVolumeMountPoint` Win32 API call and `Remove-PartitionAcessPath` should be identical to `DeleteVolumeMountPoint` Win32 API call.
I tried to call `getTarget` (with calling to `GetVolumeNameForVolumeMountPoint`) after `Add-PartitionAccessPath`, then call `DeleteVolumeMountPoint` without having issues in a local testing.

pkg/os/volume/api.go
laozc: Thanks for catching it.
I believe it was a mistake during a local merge with another commit.
I'll push a new commit to fix it.

laozc:
> @laozc do you have the link about how this WMI improves the performance? pls paste details in this PR description, thanks.

I updated the PR description to include my benchmark result mentioned in https://github.com/kubernetes-csi/csi-proxy/pull/360#issuecomment-2674022634

laozc:
@andyzhangx Good to see the change works well on azuredisk-csi-driver.
Could you check if there is any other change needed on this PR?
I will rebase other PRs later.
The current review order is
Volume (https://github.com/kubernetes-csi/csi-proxy/pull/360) => Disk (https://github.com/kubernetes-csi/csi-proxy/pull/376) => System (https://github.com/kubernetes-csi/csi-proxy/pull/375) => iSCSI (https://github.com/kubernetes-csi/csi-proxy/pull/377) => SMB (https://github.com/kubernetes-csi/csi-proxy/pull/378)
after breaking down the original PRs into per-API basis.

andyzhangx:
> @andyzhangx Good to see the change works well on azuredisk-csi-driver. Could you check if there is any other change needed on this PR? I will rebase other PRs later. The current review order is Volume (#360) => Disk (#376) => System (#375) => iSCSI (#377) => SMB (#378) after breaking down the original PRs into per-API basis.

@laozc do you have any debug tips about how to call WMI on Windows node? we use powershell command on windows node, that's quite easier in debugging issues.

pkg/cim/wmi.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+	defer outparams.Clear()
+	returnRaw, err := outparams.ToIDispatch().GetProperty("ReturnValue")
+	if err != nil {
+		return
+	}
+	defer returnRaw.Clear()
+	if returnRaw.Value() != nil {
+		result.ReturnValue = returnRaw.Value().(int32)
+		log.Printf("[WMI] - Return [%d] ", result.ReturnValue)
```
andyzhangx, line 171: klog.V(6).Infof

pkg/cim/wmi.go
```diff
@@ -0,0 +1,368 @@
...
+			return
+		}
+		defer returnRawIn.Clear()
+
+		value, err1 := cim.GetVariantValue(returnRawIn)
+		if err1 != nil {
+			err = err1
+			return
+		}
+		// log.Printf("OutParam [%s]=> [%+v]\n", outp.Name, value)
```
andyzhangx, line 187: remove

pkg/cim/wmi.go
```diff
@@ -0,0 +1,368 @@
...
+		defer inparamsRaw.Clear()
+
+		inparams, err := oleutil.CallMethod(inparamsRaw.ToIDispatch(), "SpawnInstance_")
+		if err != nil {
+			return nil, err
+		}
+		defer inparams.Clear()
+
+		for _, inp := range inParam {
+			// 	log.Printf("InParam [%s]=>[%+v]\n", inp.Name, inp.Value)
```
andyzhangx, line 149: remove

pkg/cim/wmi.go
```diff
@@ -0,0 +1,368 @@
...
+	err := QueryFromWMI(namespace, query, func(instance *cim.WmiInstance) (bool, error) {
+		instances = append(instances, instance)
+		return true, nil
+	})
+	return instances, err
+}
+
+// TODO: fix the panic in microsoft/wmi library and remove this workaround
+func executeClassMethodParam(classInst *cim.WmiInstance, method *cim.WmiMethod, inParam, outParam cim.WmiMethodParamCollection) (result *cim.WmiMethodResult, err error) {
+	log.Printf("[WMI] - Executing Method [%s]\n", method.Name)
```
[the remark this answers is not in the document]
laozc: This is a workaround to a potential issue in microsoft/wmi library.
Created https://github.com/microsoft/wmi/issues/167 to track and link the issue.

pkg/cim/wmi.go
```diff
@@ -0,0 +1,367 @@
...
+func QueryInstances(namespace string, query *query.WmiQuery) ([]*cim.WmiInstance, error) {
+	var instances []*cim.WmiInstance
+	err := QueryFromWMI(namespace, query, func(instance *cim.WmiInstance) (bool, error) {
+		instances = append(instances, instance)
+		return true, nil
+	})
+	return instances, err
+}
+
+// TODO: fix the panic in microsoft/wmi library and remove this workaround
```
andyzhangx, line 94: if this panic is only for UpdateHostStorageCache on MSFT_StorageSetting, I think we could keep the original powershell way until there is a mature solution?

laozc: Correct.
It's only impacted when the class method has no input parameters.

We'll have 4 APIs which uses CIM class method calls eventually.
1. `	result, outParams, err := InvokeCimMethod(WMINamespaceStorage, "MSFT_iSCSITarget", "Connect", inParams)`
2. `	result, _, err := InvokeCimMethod(WMINamespaceStorage, "MSFT_iSCSITargetPortal", "New", params)`
3. `	result, _, err := cim.InvokeCimMethod(cim.WMINamespaceStorage, "MSFT_StorageSetting", "UpdateHostStorageCache", nil)`
4. `	result, _, err := cim.InvokeCimMethod(cim.WMINamespaceStorage, "MSFT_iSCSISession", "SetCHAPSecret", map[string]interface{}{"ChapSecret": mutualChapSecret})`

`UpdateHostStorageCache` is the only one falling in this case.
https://learn.microsoft.com/en-us/windows-hardware/drivers/storage/msft-storagesetting-updatehoststoragecache

This is a defect in `microsoft/wmi`. Posted the analysis in https://github.com/microsoft/wmi/issues/167 and raised a [PR](https://github.com/microsoft/wmi/pull/168) to fix it in this upstream dependency.

We only need to move this line to https://github.com/kubernetes-csi/csi-proxy/pull/360/files#diff-90b6f12c56e11a97a650f55b092ccee8d80ff2e4cb2294394838160aaf39140aR228
to `method.Execute` after the change was delivered in `microsoft/wmi` module.

I believe we could use this workaround before the change in `microsoft/wmi` is merged. In the meantime, we could ask the `microsoft/wmi` maintainers to take a look on my fix. We will make a small change after the fix is merged with a version bump.

andyzhangx (approved):
/lgtm
/hold cancel

pkg/os/volume/api.go
```diff
@@ -149,92 +177,118 @@ func (VolumeAPI) UnmountVolume(volumeID, path string) error {
...
+	}
 
-		if err != nil || len(out) == 0 {
-			return fmt.Errorf("error getting sizemin,sizemax from mount. cmd: %s, output: %s, error: %v", cmd, string(out), err)
+	// If size is 0 then we will resize to the maximum size possible, otherwise just resize to size
+	if size == 0 {
+		var sizeMin, sizeMax ole.VARIANT
+		var status string
+		result, err := part.InvokeMethodWithReturn("GetSupportedSize", &sizeMin, &sizeMax, &status)
+		if result != 0 || err != nil {
```
andyzhangx, line 206: I got an unrecovered error in one e2e test:

```
could not resize volume 3 (\var\lib\kubelet\plugins\kubernetes.io\csi\disk.csi.azure.com\0bb9740d61a879a319cbd43b171f42ada33491eb75ea7198602e4a01bc5ca669\globalmount):  error getting sizemin, sizemax from volume (\\?\Volume{250aeb0d-22a0-4b29-b3e1-98d0c9c3c5b7}\). result: 4, error: <nil>
```

@laozc do you know what does `result: 4` mean?

andyzhangx: from chatGPT, 4 also means successful? (not sure whether it's correct), I think we only need to check whether err == nil or not in this case?
```
Here is a general interpretation of return values in WMI methods:

Return Value 0: The method was successful.
Return Value 4: The method was successful.
Return Value 8: The method was partially successful.
Return Value 9: The method was not successful.
```

laozc: Do you mean the test failed on `GetSupportedSize`?
Could you point me to the test case?

The call to
`result, err := part.InvokeMethodWithReturn("GetSupportedSize", &sizeMin, &sizeMax, &status)`
actually invokes
`results, err := c.InvokeMethod(methodName, params...)`
which should be the return value of this OLE call.
From the documentation, it should indicate a failure
https://learn.microsoft.com/en-us/windows-hardware/drivers/storage/msft-partition-getsupportedsizes
in general

> Success (0)
> 
> Not Supported (1)
> 
> Unspecified Error (2)
> 
> Timeout (3)
> 
> Failed (4)

andyzhangx: @laozc that's a intermittent test failure, not easy to repro, I think we need to print out the status if there is error

```
ExtendedStatus [out]

A string that contains an embedded [MSFT_StorageExtendedStatus](https://learn.microsoft.com/en-us/windows-hardware/drivers/storage/msft-storageextendedstatus) object.

This parameter allows the storage provider to return extended (implementation-specific) error information.
```

andyzhangx: this PR would provide more logs: https://github.com/kubernetes-csi/csi-proxy/pull/383

PR #362, opened 2024-12-09
Prow update master

TerryHowe:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Update release-tools according to https://github.com/kubernetes-csi/csi-release-tools/issues/7

**Which issue(s) this PR fixes**:
Related https://github.com/kubernetes-csi/csi-release-tools/pull/259
Closes: https://github.com/kubernetes-csi/csi-proxy/pull/353

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #363, opened 2024-12-11
Add min go runtime to be 1.23 and add  godebug winsymlink=0

mauriciopoppe:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

This is in response to this thread
https://groups.google.com/g/container-storage-interface-community/c/btNUUr67OMA/m/DFk-N-dBAQAJ

---
golang 1.23 has made some significant changes to how Windows mount points are reported. We are still evaluating the impact, so until we fully understand what fixes we may need to make, CSI drivers and csi-proxy should either:

1. Hold off on upgrading to golang 1.23 until we provide further guidance. This also means refraining from updating [k8s.io/mount-utils](http://k8s.io/mount-utils) to v0.32.0
2. If upgrading to golang 1.23, add the "godebug winsymlink=0" flag to the go.mod file to disable these changes.

See https://github.com/kubernetes/kubernetes/issues/129080 for more details.

---

Through https://github.com/kubernetes-csi/csi-proxy/pull/362 we used go 1.23 to build the binary so I think we should set a min go version and `godebug winsymlink=0` in go.mod

```release-note
Set min go version to 1.23 and godebug winsymlink=0 in go.mod
```

/cc @msau42

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -4,7 +4,7 @@ jobs:
   integration_tests:
     strategy:
       matrix:
-        go: ['1.22']
+        go: ['1.23']
         platform: [windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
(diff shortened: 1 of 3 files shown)


andyzhangx (approved):
/lgtm

PR #364, opened 2024-12-16
fix: unnecessary resize volume error

andyzhangx:
**What type of PR is this?**
/kind bug
fix: unnecessary resize volume error

this PR fixes the following resize volume error on Windows just right after disk format and mount succeeds on the node:
```
Warning  FailedMount             4s   kubelet                  MountVolume.MountDevice failed for volume "pvc-40e35e3f-2bf0-4adf-a27f-8a42f3a4ccf5" : rpc error: code = Internal desc = NodeStageVolume: could not resize volume 3 (\var\lib\kubelet\plugins\kubernetes.io\csi\disk.csi.azure.com\688ffc68ac8e6eb66ee127431af137461becbce570a6bfba581d3483ad7d778f\globalmount):  error resizing volume. cmd: Get-Volume -UniqueId "$Env:volumeID" | Get-Partition | Resize-Partition -Size 32195460608, output: Resize-Partition : Size Not Supported\r
 \r
Extended information:\r
The size of the extent is less than the minimum of 1MB.\r
 \r
Activity ID: {0ce86b89-e624-484f-a5aa-9810dccdf9b6}\r
At line:1 char:56\r
+ ...  "$Env:volumeID" | Get-Partition | Resize-Partition -Size 32195460608\r
+                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\r
    + CategoryInfo          : NotSpecified: (StorageWMI:ROOT/Microsoft/.../MSFT_Partition) [Resize-Partition], CimExce \r
   ption\r
    + FullyQualifiedErrorId : StorageWMI 4097,Resize-Partition\r
```

 - root cause

we use `Get-Volume -UniqueId \"$Env:volumeID\" | Get-partition | Get-PartitionSupportedSize | Select SizeMax `
to get the targetVolumeSize, and we use  following command to get the currentVolumeSize
```
(Get-Volume -UniqueId \"$Env:volumeID\\" | Get-partition).Size
```
if targetVolumeSize > currentVolumeSize, then we should resize.

But as following example, you could see for a 29.98 GB disk, just right after it's formatted, targetVolumeSize is always bigger than currentVolumeSize (just 1MB bigger), so resize operation would always happen. I think this is a bug, only if targetVolumeSize > currentVolumeSize +  1GB, then we should perform resize operation, in that case, NodeStageVolume should succeed.

```
(Get-Disk -Number 3 | Get-Partition | Get-Volume).UniqueId
PS C:\> Get-Volume -UniqueId "\\?\Volume{ccfda246-2790-4cf3-9fbf-37a62e1f9f6d}\" | Get-partition
   DiskPath: \\?\scsi#disk&ven_msft&prod_virtual_disk#6&23db5f40&0&000001#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}

PartitionNumber  DriveLetter Offset                                                           Size Type
---------------  ----------- ------                                                           ---- ----
2                G           16777216                                                     29.98 GB Basic

PS C:\> (Get-Volume -UniqueId "\\?\Volume{ccfda246-2790-4cf3-9fbf-37a62e1f9f6d}\" | Get-partition).Size
32194428928
PS C:\> Get-Volume -UniqueId "\\?\Volume{ccfda246-2790-4cf3-9fbf-37a62e1f9f6d}\" | Get-partition | Get-PartitionSupportedSize | Select SizeMax

    SizeMax
    -------
32195460608
```

Fixes #

```release-note
fix: unnecessary resize volume error
```

### DIFF
--- a/pkg/os/volume/api.go
+++ b/pkg/os/volume/api.go
@@ -197,9 +197,9 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
 		return fmt.Errorf("error getting the current size of volume (%s) with error (%v)", volumeID, err)
 	}
 
-	//if the partition's size is already the size we want this is a noop, just return
-	if currentSize >= finalSize {
-		klog.V(2).Infof("Attempted to resize volume %s to a lower size, from currentBytes=%d wantedBytes=%d", volumeID, currentSize, finalSize)
+	// only resize if finalSize - currentSize is greater than 100MB
+	if finalSize-currentSize < 100*1024*1024 {
+		klog.V(2).Infof("minimum resize difference(1GB) not met, skipping resize. volumeID=%s currentSize=%d finalSize=%d", volumeID, currentSize, finalSize)
 		return nil
 	}
 


pkg/os/volume/api.go
```diff
@@ -197,8 +197,8 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
 		return fmt.Errorf("error getting the current size of volume (%s) with error (%v)", volumeID, err)
 	}
 
-	//if the partition's size is already the size we want this is a noop, just return
-	if currentSize >= finalSize {
+	// only resize if finalSize - currentSize is greater than 100MB
+	if finalSize-currentSize < 100*1024*1024 {
```
mauriciopoppe, line 201: Is 1GB the minimal quantity allowed for resize. I think this would depend on the provider right?

mauriciopoppe: ah I think I misread it, my bad.

msau42:
Awesome, we'll also want to backport this to csi proxy v1 was well.

mauriciopoppe:
`master` is CSI Proxy v1 but agreed this should go against the branch https://github.com/kubernetes-csi/csi-proxy/tree/library-development as well.

andyzhangx:
> Is the description correct? The implementation says 100MB in a comment but 1GB in the description.

delta of the minimum size expand should be at least 1GB, so 100MB is ok.

pkg/os/volume/api.go
```diff
@@ -197,8 +197,8 @@ func (VolumeAPI) ResizeVolume(volumeID string, size int64) error {
...
 		klog.V(2).Infof("Attempted to resize volume %s to a lower size, from currentBytes=%d wantedBytes=%d", volumeID, currentSize, finalSize)
```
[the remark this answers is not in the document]
andyzhangx: fixed

PR #368, opened 2025-01-06
Fix format error in log message

mauriciopoppe:
**What type of PR is this?**
>

kind bug

Fix format error in log message

```release-note
NONE
```

/cc @andyzhangx


andyzhangx (approved):
/lgtm

Issue #361, opened 2024-12-05
Sweep and adjust filesystem Stat/Lstat/EvalSymlinks calls in go 1.23


mauriciopoppe:
Context: https://github.com/kubernetes/kubernetes/issues/129084

Go 1.23 changed stdlib behavior of filesystem calls Stat / Lstat / EvalSymlinks on Windows. This broke some kubelet handling of volumes on Windows, and possibly other use of those functions. For Kubernetes 1.32, the behavior was temporarily reverted via godebug switches in https://github.com/kubernetes/kubernetes/pull/129083, but this is not a long-term solution. Libraries like [k8s.io/mount-utils](https://github.com/kubernetes/mount-utils) can be used by downstream consumers building with go 1.23 and get unintented behavior.

Even though this project hasn't been updated to use go 1.23 (still using go 1.20), it's possible that the project is built from source using a different go runtime. If the project is built with go 1.23+

- The go.mod file should have these entries https://github.com/kubernetes/kubernetes/issues/129080#issuecomment-2517583557 OR
- The filesystem calls Stat / Lstat / EvalSymlinks should be analyzed in the CSI Proxy codebase to make sure that they're compatible with the go 1.23 changes and if needed adjust them to work properly.

Also read https://github.com/kubernetes/kubernetes/pull/129103

liggitt:
note that we *may* just be setting the `winsymlink` option, see discussion in https://github.com/kubernetes/kubernetes/pull/129103

laozc:
/close
this is no longer needed after we moved to Windows API.

PR #371, opened 2025-02-21
Set up a metrics server with data served on /metrics

laozc:
**What type of PR is this?**

/kind feature

Expose gRPC prometheus metrics to capture performance for volume operations 

Fixes #

```release-note
Added flag metrics-bind-addr to specify the endpoint for gRPC server metrics
```


pkg/server/server.go
```diff
@@ -70,6 +84,24 @@ func (s *Server) startListening() (chan *versionedAPIDone, []error) {
 	return s.createAndStartGRPCServers(listeners), nil
 }
 
+//
+//func (s *Server) createOtelExporter(ctx context.Context) {
```
mauriciopoppe, line 88: Please remove unused code.

mauriciopoppe (commented):
Thanks for the PR, I left a few comments.

cmd/csi-proxy/main.go
```diff
@@ -68,7 +72,22 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	s := server.NewServer(apiGroups...)
+
+	reg := prometheus.NewRegistry()
+	go func() error {
+		m := http.NewServeMux()
+		// Create HTTP handler for Prometheus metrics.
+		m.Handle("/metrics", promhttp.HandlerFor(
```
mauriciopoppe, line 80: It seems that there's a generic impl in the SMB CSI Driver 

https://github.com/kubernetes-csi/csi-driver-smb/blob/a71b8cf5b868091fdaa6f52842a1e8ae9fe6c98e/cmd/smbplugin/main.go#L122

Could we use https://github.com/kubernetes/component-base/tree/master/metrics too?

go.mod
```diff
@@ -1,37 +1,58 @@
 module github.com/kubernetes-csi/csi-proxy
 
-go 1.23
```
mauriciopoppe, line 3: This project became incompatible with go1.23 and above because of https://github.com/kubernetes-csi/csi-proxy/issues/361. Please rebase.

cmd/csi-proxy/main.go
```diff
@@ -34,11 +36,12 @@ func (i *workingDirFlags) Set(value string) error {
...
 var (
-	kubeletPath    = flag.String("kubelet-path", `C:\var\lib\kubelet`, "Prefix path of the kubelet directory in the host file system")
-	windowsSvc     = flag.Bool("windows-service", false, "Configure as a Windows Service")
-	requirePrivacy = flag.Bool("require-privacy", true, "If true, New-SmbGlobalMapping will be called with -RequirePrivacy $true")
-	service        *handler
-	workingDirs    workingDirFlags
+	kubeletPath     = flag.String("kubelet-path", `C:\var\lib\kubelet`, "Prefix path of the kubelet directory in the host file system")
+	windowsSvc      = flag.Bool("windows-service", false, "Configure as a Windows Service")
+	requirePrivacy  = flag.Bool("require-privacy", true, "If true, New-SmbGlobalMapping will be called with -RequirePrivacy $true")
+	metricsBindAddr = flag.String("metrics-bind-addr", "localhost:10010", "The address the metric endpoint binds to.")
```
mauriciopoppe, line 42: I see that kube-proxy uses `metrics-bind-address`, let's use that as well. 

https://github.com/kubernetes/kubernetes/blob/dc3021b156a6a5adc4b4898d366822529bcc74a4/cmd/kube-proxy/app/options.go#L119

mauriciopoppe, line 42: What should be a good default for the port? I think we should pick a default in a way that doesn't conflict with any other process that's running in the node.

pkg/server/server.go
```diff
@@ -107,8 +119,25 @@ func (s *Server) createAndStartGRPCServers(listeners []net.Listener) chan *versi
 	doneChan := make(chan *versionedAPIDone, len(s.versionedAPIs))
 	s.grpcServers = make([]*grpc.Server, len(s.versionedAPIs))
 
+	//s.createOtelExporter(context.Background())
+
 	for i, versionedAPI := range s.versionedAPIs {
-		grpcServer := grpc.NewServer()
+		opts := []grpc.ServerOption{
```
mauriciopoppe, line 125: Please remove unused middlewares.

pkg/server/server.go
```diff
@@ -107,8 +119,25 @@ func (s *Server) createAndStartGRPCServers(listeners []net.Listener) chan *versi
 	doneChan := make(chan *versionedAPIDone, len(s.versionedAPIs))
 	s.grpcServers = make([]*grpc.Server, len(s.versionedAPIs))
 
+	//s.createOtelExporter(context.Background())
```
mauriciopoppe, line 122: Please remove unused code.

pkg/server/server.go
```diff
@@ -28,9 +32,17 @@ func NewServer(apiGroups ...srvtypes.APIGroup) *Server {
 		versionedAPIs = append(versionedAPIs, apiGroup.VersionedAPIs()...)
 	}
 
+	srvMetrics := grpcprom.NewServerMetrics(
```
mauriciopoppe, line 35: Here's an example of setting metrics through component-base/metrics, would it be possible to reuse that code? 

https://github.com/kubernetes/kubernetes/blob/dc3021b156a6a5adc4b4898d366822529bcc74a4/pkg/scheduler/metrics/metrics.go#L149

mauriciopoppe (commented):
I left some comments.

/ok-to-test

pkg/server/server.go
laozc: Sure.

pkg/server/server.go
laozc: Sure.

cmd/csi-proxy/main.go
laozc: I changed the defaults to empty to disable the feature, which avoid introducing any user interface change.

pkg/server/server.go
laozc: Sure. Moved all the metrics related codes to a single file.

pkg/server/metrics/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+	)
+
+	metricsList = []prometheus.Collector{
+		grpcServerMetrics,
+	}
+}
+
+// Register registers a list of metrics.
+func Register() {
+	registerMetrics.Do(func() {
```
mauriciopoppe, line 33: In which scenario would `Register` be called multiple times?

pkg/server/metrics/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+		initMetrics()
+		for _, metric := range metricsList {
+			legacyregistry.RawMustRegister(metric)
+		}
+	})
+}
+
+func GRPCMetricsOptions() []grpc.ServerOption {
```
mauriciopoppe, line 41: Nit: Exported functions must be documented.

pkg/server/metrics/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+
+	metricsList = []prometheus.Collector{
+		grpcServerMetrics,
+	}
+}
+
+// Register registers a list of metrics.
+func Register() {
+	registerMetrics.Do(func() {
+		initMetrics()
```
mauriciopoppe, line 34: I'd inline the contents of `initMetrics` here

integrationtests/utils.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // startServer starts the proxy's GRPC servers, and returns a function to shut them down when done with testing
 func startServer(t *testing.T, apiGroups ...srvtypes.APIGroup) func() {
```
mauriciopoppe, line 29: Would it be possible to add an integration test that makes a request to the new `metrics/` endpoint?

pkg/server/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	metricsList = []prometheus.Collector{
+		grpcServerMetrics,
+	}
+
+	for _, metric := range metricsList {
+		legacyregistry.RawMustRegister(metric)
+	}
+}
+
+func ExportMetrics(metricsAddress string) {
```
mauriciopoppe, line 38: Nit: Exported function should be documented.

cmd/csi-proxy/main.go
```diff
@@ -68,7 +70,12 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	s := server.NewServer(apiGroups...)
+
+	enableMetrics := *metricsBindAddr != ""
+	if enableMetrics {
+		metrics.ExportMetrics(*metricsBindAddr)
```
mauriciopoppe, line 76: If I want to enable metrics and by some reason `ExportMetrics` fails then this not fail and continue silently.

What about `metrics.ExportMetrics` returning an error that's checked while setting up CSI Proxy? I think we should panic with the error returned.

pkg/server/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	if metricsAddress == "" {
+		return
+	}
+	l, err := net.Listen("tcp", metricsAddress)
+	if err != nil {
+		klog.Warningf("failed to get listener for metrics endpoint: %v", err)
```
mauriciopoppe, line 44: Please return an error here.

pkg/server/metrics/metrics.go
mauriciopoppe, line 38: This method is setting up a metrics server, what about naming it as such? e.g. `SetupMetricsServer`

pkg/server/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+		return
+	}
+	serve(l, serveMetrics)
```
mauriciopoppe, line 47: Do we need to have an additional function? What about inlining the contents of `serve` here? Same for inlining `serveMetrics`

mauriciopoppe (commented):
I wrote a few comments in the setup of the server and on inlining the code. The approach looks good to me, thanks.

mauriciopoppe:
Thanks a lot for the implementation! I thought we could port the implementation to the library-development branch but I think it's not applicable because the gRPC metrics would already be exposed as part of the CSI Driver metrics (once the CSI Driver uses the library implementation).

/lgtm
/approve

PR #372, opened 2025-02-24
Use WMI instead of PowerShell for OS operations

### LINKED ISSUE #193: High cpu usage of powershell processes triggered by csi-proxy

laozc:
**What type of PR is this?**
/kind feature

This PR leverages of WMI interfaces to replace the PowerShell functions for Windows related storage operations,
which improve the overall performance of csi-proxy.

Fixes #193

```release-note

```


mauriciopoppe:
Thanks a lot for porting the codebase to the library version! For the changes in the library version would they happen after the binary changes?

andyzhangx (approved):
/lgtm

.github/workflows/windows.yml
```diff
@@ -22,11 +22,11 @@ jobs:
           Install-WindowsFeature -name Hyper-V-PowerShell
 
           $env:CSI_PROXY_GH_ACTIONS="TRUE"
-          go test --timeout 20m -v -race ./integrationtests/...
```
mauriciopoppe, line 25: Is there a reason `-race` was removed?

laozc: Re-enabled `-race` but with `checkptr=0` disabled as part of https://github.com/kubernetes-csi/csi-proxy/pull/385
This should be related to dereferencing the pointer allocated by COM is not considered as safe in Go.
We either need to use C/CGO on these pointers or set `checkptr=0`.

PR #400, opened 2025-07-29
cleanup: Bump microsoft/wmi to 0.34.0 containing WMI method call fix (library version)

### LINKED ISSUE #193: High cpu usage of powershell processes triggered by csi-proxy

laozc:
**What type of PR is this?**
/kind cleanup

The WMI method call fix is released with microsoft/wmi @ 0.31.2.
This PR bumps this library and remove the workaround.

Fixes #193 

Refer to https://github.com/kubernetes-csi/csi-proxy/pull/394 for the change for standalone process.

```release-note
NONE
```


go.mod
```diff
@@ -1,13 +1,17 @@
 module github.com/kubernetes-csi/csi-proxy/v2
 
-go 1.22
+go 1.24.3
+
+godebug winsymlink=0
```
mauriciopoppe, line 5: Would this setting be applied transitively by CSI Drivers that import the CSI Proxy library? I think it's worth checking that with your CSI Driver given that the command to build the go binary would be in that codebase and not on this one.

go.mod
```diff
@@ -1,13 +1,17 @@
 module github.com/kubernetes-csi/csi-proxy/v2
 
-go 1.22
+go 1.24.3
```
mauriciopoppe, line 3: Agreed that we should target the version that kubernetes/kubernetes is targetting, and [it's targeting 1.24.5](https://github.com/kubernetes/kubernetes/blob/master/.go-version) so this is good.

No action items, just a comment to clarify why it's a good version.

go.mod
mauriciopoppe: One long term goal is to address https://github.com/kubernetes-csi/csi-proxy/issues/374 where we don't need the `godebug ...` rule anymore given that it's a hack to make existing code work with go 1.23+. We should ideally remove this line while ensuring that there's no regression on CSI Drivers that use this go module.

andyzhangx: check the details here: https://github.com/kubernetes/kubernetes/pull/129368, it only impacts symbolic link for disk

